PR #89, opened 2014-09-01
news on .po fr_fr

yolateng0:
A small stake in the project.


lwindolf:
Closing this very old PR. I'm sorry I actually do not remember what the issue was here. But I believe merging doesn't make sense anymore.

PR #137, opened 2015-01-13
Vertically expand the source selector

6ahodir:
Resizing the source selector window will also expand the scrolled window. Without this change the scrolled window remains the same height no matter how tall the window is.


lwindolf:
Thanks for the fix!

PR #139, opened 2015-01-14
Expose enums to GIR

6ahodir:
Hi Lars, I'm not sure if that's the right way of doing it.


lwindolf:
Looks good. Thanks for the fix!

PR #145, opened 2015-02-11
Fixes a copy&paste bug on shell item action entry

mozbugbox:


lwindolf:
Thanks a lot for this fix!

lwindolf:
Merged against 1.10 too...

PR #149, opened 2015-02-14
Fixes a random crash on startup

mozbugbox:
The `count` in feed_list_node.c/feed_list_node_update() could get
uninitialized when `labeltype` contain garbage.


mozbugbox:
Actually, if the `labeltype` initialized to 0, the `count` would still be uninitialized. It could happen if `node->capabilities` had neither show unread count nor show item count set.

Somehow the `default` branch of the `switch (labeltype)` need to have something set for the `count` value.

mozbugbox:
While we're at it, the `countColor` is leaking in that function.

lwindolf:
Thank you for the fix again.

About the uninitialized count: I could imaging just putting the default case to one of both cases. That solves the issue,

About countColor leaking: well technically yes, but it is a static var, so just no cleanup...

PR #151, opened 2015-02-15
Updating sq translation

ujdhesa:
Hello,

Here an update for sq.po translation file. I updated the translation on Translation Project as well, but there the tip there has stopped at 1.8.something.


ujdhesa:
...at 1.10-rc4 actually. That’s some 35 strings behind the master here.

lwindolf:
Thanks for the update!

PR #157, opened 2015-02-27
Update French Translation

filorin:
*Hi Lars,*

I updated French Translation according to French writing rules we already use in GNOME.
So, what did i do ?
* Replace the space char by Non-breaking space where needed.
* Replace " by « or » where needed
* Corrected some problems that generated too large windows because of too long sentences.

PS : i will use github instead of using the Translation Project for future commits.

*Regards,
Guillaume*


lwindolf:
Thanks for the update.

PR #162, opened 2015-03-08
Use header for important part of the README

asl97:
so that we could give link like:
https://github.com/lwindolf/liferea#building-liferea-yourself
https://github.com/lwindolf/liferea/blob/master/README.md#introduction
https://github.com/asl97/liferea/tree/patch-3#debugging-crashes

and **after** this is merge:
https://github.com/lwindolf/liferea#debugging-crashes

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,14 +27,14 @@ any problems compiling the packages or from GIT don't hesitate to
 contact us in IRC (see Support section) to help you with it!
 
 
-_Dependencies_
+###### _Dependencies_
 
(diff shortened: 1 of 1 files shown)


asl97:
ah.... github use pjax and partial change,  clicking the link on the same repo doesn't seem to auto scroll to the line...

never mind,  you can reopen this and merge it if you want but i will close this for now.

lwindolf:
It is a nice change. So merging it :-)

PR #163, opened 2015-03-13
Gir browser tab

mozbugbox:
Add gir introspect for browser tabs so we can monitor the creation of new htmlviews


lwindolf:
Nice refactoring!

PR #168, opened 2015-03-30
Release Peas plugin before release widgets

mozbugbox:
The widgets might be referred by deactivate calls in plugins


lwindolf:
Thanks!

PR #169, opened 2015-04-01
Fix of date format with 2 digit year never reached

one-more-alex:
4 digit year sucseeded with negative tm.tm_year.
For example, if date is "Wed, 01 Apr 15 11:39:23 +0300", then tm_year became negative.
Date format taken from http://vesti-ukr.com/feed/53-glavnye-vesti-strany.rss.


lwindolf:
Great debugging. Thank you for contributing this fix!

PR #174, opened 2015-04-14
Consider <base> tag when discover feeds

mozbugbox:


lwindolf:
Awesome! Thanks for this!

PR #178, opened 2015-04-17
Fullscreen mode

mozbugbox:
This pull request implement fullscreen mode. I'm not sure if the implementation is clean enough. It works for me!


lwindolf:
Awesome! I'll test it thouroughly, the implementation so far looks very good. Now watching movie trailers is even more fun.

PR #185, opened 2015-05-05
Update Spanish translation

fitojb:
Hi Lars, here is the update. I want to apologize if I sounded rude before, it wasn’t my intention.


lwindolf:
Thank you for the improvements!!!

@fitojb Don't worry about the discussion. I found it constructive and ok.

PR #196, opened 2015-06-01
Updated Russian

bboa:
Thanks for nice app


lwindolf:
Thank you for the update!

PR #210, opened 2015-07-05
Enable tests for parsing RFC822 dates with two digit years

arunanbala:
This commit just turns on two pre-existing tests and adds another case. The tests were not previously included as they did not pass when written. The date parser has since been fixed to make them work. An extra case was added to check the particular feed that first got me looking at this.


lwindolf:
Awesome. Thanks for updating the tests!

PR #227, opened 2015-08-03
Fix dialog ui load with latest gtkbuilder

yaneti:
With recent gtk3 loading the feed preferences dialog fails with:
/usr/share/liferea/liferea.ui:519:65 Object with ID liststore6 not found
Add liststore6 to the explicit objects loaded.


yaneti:
The error message can only be seen if adding a GError to the gtk builder call. 
With the current call its just an g_error abort.

yaneti:
btw, I see that you've tagged this of 1.11.6.
I observed this on rawhide with 1.10.16 so filed the pull request for liferea-1_10 . Didn't have the chance to see if master was affected...

lwindolf:
Sorry, I was a little confused. Tag is fixed now.

Issue #256, opened 2015-10-02
PyGIWarnings on loading plugins


glitsj16:
Getting some PyGIWarnings when loading plugins, related to importing without specifying a version first.
/usr/lib/liferea/plugins/trayicon.py:22: PyGIWarning: Peas was imported without specifying a version first. Use gi.require_version('Peas', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import GObject, Peas, PeasGtk, Gtk, Liferea, Gdk
/usr/lib/liferea/plugins/trayicon.py:22: PyGIWarning: PeasGtk was imported without specifying a version first. Use gi.require_version('PeasGtk', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import GObject, Peas, PeasGtk, Gtk, Liferea, Gdk
/usr/lib/liferea/plugins/trayicon.py:22: PyGIWarning: Liferea was imported without specifying a version first. Use gi.require_version('Liferea', '3.0') before import to ensure that the right version gets loaded:

  from gi.repository import GObject, Peas, PeasGtk, Gtk, Liferea, Gdk
/usr/lib/liferea/plugins/gnome-keyring.py:22: PyGIWarning: GnomeKeyring was imported without specifying a version first. Use gi.require_version('GnomeKeyring', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import GObject, Peas, PeasGtk, Gtk, Liferea, GnomeKeyring
/usr/lib/liferea/plugins/media-player.py:26: PyGIWarning: Gst was imported without specifying a version first. Use gi.require_version('Gst', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import GObject, Peas, PeasGtk, GLib, Gtk, Liferea, Gst

 These are non-fatal warnings, but seem to be easy fixes. I'm forking at the moment and will add a pull request to properly silence these warnings. I'm far from an experienced developer though, so I might not see the whole picture regarding implications for liferea. Please advise or disregard the coming pull request :-)

---
arch linux, liferea --> 1.11.5.r5336.g2f3d1aa-1

PR #257, opened 2015-10-02
Fix PyGIWarnings on plugins

glitsj16:
This fixes GitHub #256


lwindolf:
Thank you for this contribution!

PR #262, opened 2015-10-07
Fix additional PyGIWarnings on media-player plugin

glitsj16:
This should fix all the PyGIWarnings on the plugins.


lwindolf:
Thanks again!

Issue #176, opened 2015-04-15
Any plans to switch to GTK3 Header Bars?


jurf:
It would save a lot of screen space and give a Liferea a fresh new look.

Anyway, just an idea.

lwindolf:
We will probably migrate some day. Just found no time to convert the code. I estimate it needs a lot of reorganization.

DoctorJellyface:
Great, thanks!

gusnan:
PLEASE make this optional. For me (and I belive many others who doesn't use Gnome3), the new GTK3 stuff is simply a no-go.

asl97:
@gusnan it already recommenced for liferea to be build against gtk3, in fact, gtk3 is listed under mandatory.

https://github.com/lwindolf/liferea#dependencies
https://github.com/lwindolf/liferea#mandatory

also, iirc, there are lesser and lesser distro being ship with Gnome2 and
GTK3 isn't that new anymore (it was release in 2011).

gusnan:
Does that mean it _requires_ the new style of titlebars?

DoctorJellyface:
No, but AFAIK client side decorations have support with almost every window manager out there. Wayland even requires CSD.

lwindolf:
@gusnan In the scope of Liferea (one man project) supporting two variants isn't possible.

gusnan:
Well, then the logical thing would be to support the alternative that requires the least changes, and not a redesign. But, if a majority want the new stuff, then I'll simply will have to accept it.

lwindolf:
@gusnan Don't worry GNOME 4 will have even more "new" stuff for sure! ;-)

gnumdk:
On demand headerbar is easy, here what i've done in lollypop (python code)
```python
        toolbar = Gtk.HeaderBar()
        # Only set headerbar if according DE detected or forced manually
        if use_csd():
            self.set_titlebar(toolbar)
            self._toolbar.set_show_close_button(True)
            self.add(main_widget())
        else:
            hgrid = Gtk.Grid()
            hgrid.set_orientation(Gtk.Orientation.VERTICAL)
            hgrid.add(toolbar)
            hgrid.add(main_widget)
            hgrid.show()
            self.add(hgrid)
```
Most of the work is about redesign app to get ride of menus ;)

tpo:
FWIW - I also wouldn't welcome the GTK3 header bars, because:

[...]

PR #276, opened 2015-10-23
Gtk 3 header bar for some desktop environments

Leiaz:
I replaced the GtkAction that are being deprecated by GAction, in order to make a new style "application menu". Until the last two commits, I just recreated the menubar and toolbar with the GActions.

I wanted to see what would be required to have the header bar discussed in issue #176.

The last two commits make the application use GtkHeaderBar as titlebar (enables "client side decoration") when the desktop environment "prefers" an app menu (determined by [gtk_application_prefers_app_menu](https://developer.gnome.org/gtk3/stable/GtkApplication.html#gtk-application-prefers-app-menu)), the "old style" toolbar and menubar otherwise.

I don't like how the "client side decoration" ignores the window manager. It doesn't work well in XMonad, and probably other tiling or minimal window managers. But those don't "prefers an app menu".

I don't know Gtk very well, so feel free to say if my code is horrible. :)


Leiaz:
The header bar requires Gtk 3.10. The functions Travis can't find are new to 3.10.

lwindolf:
@Leiaz Awesome change. Quite large so I need 2-3 days to check it. Please have patience :-)

Gert-dev:
Don't want to sound like the guy that asks for a status report every so often - I understand contributions are voluntary and time is precious -, but is this still on the roadmap? I feel a header bar and perhaps some other small visual changes such as use of the new GTK 3 widgets could do some real good to Liferea's general feel and make it feel a little more modern.

lwindolf:
@Gert-dev I pains me to have this contribution of Leiaz waiting here so long. I also agree that the fresh look by the header bars would benefit the image of Liferea.

Around last fall I was close to dropping Liferea altogether, mostly due to the continuosly bad balance of average motivation and GTK/GNOME release cycles triggering endless refactoring work. The continuous visual changes to catch up with the Apple/Win10 styles also eat up a lot of my motivation. Actually this changes is such a change.

From a pure versioning perspective this belongs into 1.13 and needs to be optional. Judging from the trayicon situation this will become as untestable and a support pain.

@Leiaz Sorry, about me being so open on this!

lwindolf:
Further thinking about it

Gert-dev:
I understand your frustration; I can only guess at how it feels for you on such a long term and large open source project such as liferea.

Of course liferea functionally works perfectly fine for daily usage and, regardless of community misinformation regarding lack of backwards compatibility in GTK itself, has worked well for a long time, despite it not being a pinnacle of GNOME's GTK3 standards.

There are, of course, newer feed readers such as GNOME News and Feedreader that do look visually more appealing, but neither come anywhere close to liferea in terms of functionality - in my case, it is the custom conversion filter support in particular that I haven't found anywhere else yet, save for hosting your own HTTP server to do it.

This is also why I posted here; people are quickly drawn to applications that look visually interesting and appealing - it is usually perceived as new or modern, even if the code base isn't always great. By just switching the titlebar with a header bar - even if there are just a few buttons in it that contain the same menu items as are currently visible in the other bars - gives the feeling that liferea is moving along with the rest of the GNOME suite. It may - but don't shoot me if it doesn't 😄 - also attract more contributors from the GNOME community.

I understand that your first incentive is to make this optional or in a plugin. I don't know the internals of liferea, but I'd be wary of this as it may make your life - or the life of the person that has to maintain the plugin - a lot harder as you now have two interfaces to maintain and sooner or later, at least if you aren't planning on rewriting liferea in Qt or another toolkit, you'll have to do it anyway (that doesn't mean you can't wait a couple of years before doing a larger rewrite, though). It may be a good idea to just select the route that you think is best and just speed off in that direction. You can always reiterate later.

In the end, regardless of the choice you make, someone will always complain - and they have a right to, but they don't have entitlement to expect you to maintain support for three different GUI's. I think the community would rather have a supported liferea, albeit not entirely visually what they want, then a dead liferea that had an identity crisis before it died 😄.

Finally, you mentioned you thought about dropping liferea altogether. This is not a bad thing to say and if maintaining liferea is taking too much out of you, there is no shame in putting up a note in the README that you're looking for a new core maintainer. If it turns out no one wants to take it over, then so be it, this is not your fault. Change sometimes has a way of happening only when the situation is the most dire.

Leiaz:
> @Leiaz Sorry, about me being so open on this!

Why ? I definitely don't mind you being more open :)

For instance, if you don't want me to bother you with a pull request for small bug fixes like #561, please say so. You gave me Collaborator permissions but I just assumed you prefer to review everything that goes in master. Or if there is specific stuff you want me to do, don't hesitate to ask.

Some years ago, I looked at other feed readers but came back to Liferea, it is just more complete, despite it being broken on Arch Linux at the time. And it can become even better. The HTML5 enrichment feature needs some polish, but it makes some feeds actually useful :)

> The continuous visual changes to catch up with the Apple/Win10 styles also eat up a lot of my motivation. Actually this change is such a refactoring.

I'm not a fan of the new Gnome style, or of "modern" style apps for that matter. I also use Claws Mail. I prefer more features in an awkward ui to the shiny new ui where you have to guess which button hides a menu.

I don't think the way I kept both toolbar and header bar working above, cramming the toolbar in the header bar, would be Gnome style compliant :)

But the GtkAction to GAction change is not a visual thing, this could be done while keeping the current appearance. The GtkActions have been deprecated for a while, so I was nervous that a new Gtk 3 version would just break those actions and us Arch Linux users would end up with unusable menus and buttons :D But thankfully, it didn't happen, and now won't happen before Gtk 4, so we should have time to prepare :).

Multiple styles of menus may not be too hard to support, as they would all use the same actions.

I'll look into rebasing the GAction commits and see if custom menus and bars can easily be done with plugins. (I agree with @Gert-dev that using plugins doesn't mean we get rid of the support problem, people might not like when we break their plugins :))

lwindolf:
Guys, I create a small (not yet really functional plugin) which makes Liferea look like this:

bildschirmfoto von 2018-01-29 22-40-45

Aside from the need to switch all stuff to GAction and GtkBuilder (I believe those changes in this PR should be merged in any case) the plugin itself gives a nice isolated implementation and maintainers have the freedom to opt-in or not (by changing the default of the "active-plugins" dconf key). In my opinion this 100% solves the support hell and allows releasing this with 1.12. Later on we can make the plugin 

- on per-default, but optional
- mandatory and hidden
- or hard-code the behaviour in C

What do you think?

Leiaz:
Nice. I think we need #495 to go with it, just in case. Or at least for testing :D 

For the future, it depends if it is far enough into the future. For now I don't like the bar because any click on it that includes a small involuntary drag is interpreted as an attempt to move it. I am using xmonad and this is not supported so the window "jumps" in a corner.

In a future where it can be made to behave and not try to manage the window and/or where we have all switched to Wayland : most actions in Liferea's menus are better used from the context menus, and the rest could be put in the bar as one or two menus and buttons, so we wouldn't lose much by making it the hard coded default ... Perhaps it could have the current article title as a subtitle, to look less like a waste of space :)

Also we'll have to see how well it works as a plugin : for now it causes this warning `(liferea:4160): Gtk-WARNING **: gtk_window_set_titlebar() called on a realized window`, but it still looks like "it works".

lwindolf:
In the future: given the current release cycle (>1y) and making it mandatory, being a major change would mean 1.14.

The "realized window" warning cannot be avoided as plugins do depend on a visible window :-(

I agree about #495 

BTW about you being project collaborator: I believe your code quality to be much higher than my own, so there is no danger of you pushing without review. Don't worry. I only do care about extra code complexity. If you are interested I can also switch to feature branches for the stuff I write and perform reviews. That would benefit the code a lot. Let me know what you think!

Leiaz:
Replaced by #608

PR #277, opened 2015-10-23
libnotify plugin : disconnect the signal when deactivating

Leiaz:
This disconnects the signal when deactivating the plugin, so notifications will stop, and not multiply when the plugin is deactivated and reactivated.


lwindolf:
Thank you very much for this improvement!

PR #279, opened 2015-10-24
Fixes rules not visible in searchdialog

Leiaz:
The rules are not visible in the search dialog. It has the same problem the search folder creation dialog had.

This commit :
Moves searchdialog in its own ui file. 
Replaces deprecated GtkHBox and GtkVBox by GtkBox. 
Sets box to expand so the rules are visible.
Changes default width to 600 (like the search folder dialog) so the rules don't have scrollbars by default.


lwindolf:
I backported this to 1.10 too and added the new search.ui to po/POTFILES.in so translations do not get broken.

PR #292, opened 2015-11-29
Trayicon plugin: show new item count

mozbugbox:


lwindolf:
I didn't want to merge all the time because the new-count is not optional (like it was in the past).

But as it contains the scaling fix that currently affects so many users I think it's worth enforcing the counter.

@mozbugbox  Thank you for the great work on the plugin!!!

PR #296, opened 2015-12-20
updated german translation

pseyfert:
Updated German translation and getting existing strings closer to [German Gnome translation guidlines](https://wiki.gnome.org/de/UebersetzungsRichtlinien)


lwindolf:
Danke für das Update!

PR #297, opened 2015-12-21
trayicon plugin: minimize to systray on window close

hugoArregui:


genodeftest:
Please note that there is no GtkStatusIcon on Gnome+Wayland, so this plugin will be broken then.

hugoArregui:
Hi, I understand the problem, but the plugin is already using GtkStatusIcon, I just intercept the window close event, to minimize the app instead of just shutting it down

asl97:
@genodeftest a new plugin can always be created for Wayland and whatever new Display Server they manage to come up with.
same goes for each and every Window Manager if and when someone want them enough to code it themselves.

this pull request is about adding a way to stop liferea from shutting down as long as there is still a plugin 'holding' on to the 'shutdown flag'

lwindolf:
Thank you for the PR! Works very well. Good idea with the delete event handler check.

hugoArregui:
Thank you very much @lwindolf, I really wanted to get this merged

PR #298, opened 2015-12-22
Prevent node-updated event in case of non sucess http status

hugoArregui:
A simple way to reproduce the problem of not checking the http status is this: add a new feed source with n elements, then node->newCount == n. Now suppose in the next update the feed doesn't change, the newCount still will be n, so the node-updated event will be fired (and if you are using libnotify plugin, a notification will appear). Using this patch, the http status 304 will prevent a lot of this scenarios.


lwindolf:
Good idea!

PR #300, opened 2015-12-26
Update manpage

GreenLunar:
Minor updates to manpage.


lwindolf:
Thanks for the update!

PR #308, opened 2016-01-15
Fixes #295 : handle protocol-relative link in item

### LINKED ISSUE #295: schema-less links in <item><link> fields are not read correctly
Some feeds (e.g. [example](http://blog.merovius.de/rss.xml)) use schema-less links. Opening these in liferea doesn't only add "http" but adds the entire siteurl (e.g. [liferea-link](http://blog.merovius.de//blog.merovius.de//2014/05/06/python-fnord-of-the-day.html) instead of [correct link](http://blog.merovius.de/2014/05/06/python-fnord-of-the-day.html)).

tried with 1.10.12 and with the master version from git.

Leiaz:


lwindolf:
Looks like a safe solution. Thanks!

PR #309, opened 2016-01-31
Update ja.po

iwaim:


lwindolf:
Thank you for this update of the Japanese translation!

PR #312, opened 2016-02-06
don't update feed on start if the autoupdate is disabled

hugoArregui:
Hi, 

the title is pretty self explanatory I think, but I should mention: this works for me, but I'm not sure if I should check interval only for -2, I'm confused by the comment in subscription.c:subscription_set_update_interval

```
    interval = -1;  /* This is evil, I know, but when this method
               is called to set the update interval to 0
               we mean "never updating". The updating logic
               expects -1 for "never updating" and 0 for
               updating according to the global update
               interval... */
```

It seems I should check for -1 instead, but it's not the case

### DIFF
--- a/src/node.c
+++ b/src/node.c
@@ -130,14 +130,19 @@ node_set_subscription (nodePtr node, subscriptionPtr subscription)
 void
 node_update_subscription (nodePtr node, gpointer user_data) 
 {
+	gint interval;
 	if (node->source->root == node) {
 		node_source_update (node);
 		return;
 	}
 	
-	if (node->subscription)
+	if (node->subscription) {
+	    interval = subscription_get_update_interval (node->subscription);
+	    if (-2 != interval) {
 		subscription_update (node->subscription, GPOINTER_TO_UINT (user_data));
-		
+	    }
+	}
+
 	node_foreach_child_data (node, node_update_subscription, user_data);
 }
 


lwindolf:
@hugoArregui The change might seem to work, but IMO isn't correct. The idea in subscription.h is that subscription_update() unconditionally updates the feed (e.g. upon user request), while subscription_auto_update() conditionally performs updates by calling subscription_update(). Therefore no interval logic belongs into subscription_update().

The initial update on startup should only be triggered by subscription_auto_update() which has the logic you want already:

        if (-1 == interval)
                conf_get_int_value (DEFAULT_UPDATE_INTERVAL, &interval);

        if (-2 >= interval || 0 == interval)
                return;         /* don't update this subscription */

The call chain on startup is like this:

1. feedlist_update()
2. node_auto_update_subscription()
3. subscription_auto_update()

I tested this by setting a non-zero update interval and "no auto update" preference and then started with a gdb break point set on subscription_update().

Are you testing with "normal" feeds or an online aggregator account (like TinyTinyRSS)?

hugoArregui:
I'm testing with normal feeds

lwindolf:
That's weird. Can you help me by running like this?

     gdb src/liferea

Then in gdb prompt enter

     break subscription_update

and start with

     run

once the break point triggers:

     bt

and post the back trace here?

hugoArregui:
hey @lwindolf something I didn't clarify (sorry, I should have done it), I can reproduce the problem when I disable the auto update in the properties of a feed, I'm not talking about the general option.

Anyway, I'll send you the output you requested.

hugoArregui:
So this is the output you requested, liferea has only one feed, and the autoupdate for that feed is disabled.

```
Thread 1 "liferea" hit Breakpoint 3, subscription_update (subscription=0xf43650, flags=flags@entry=0) at subscription.c:266
266		if (!subscription)
(gdb) bt
#0  subscription_update (subscription=0xf43650, flags=flags@entry=0) at subscription.c:266
#1  0x000000000042aca4 in node_update_subscription (node=0xf43990, user_data=0x0) at node.c:139
#2  0x000000000042ac38 in node_foreach_child_full (node=<optimized out>, func=0x42ac80 <node_update_subscription>, params=1, 
    user_data=0x0) at node.c:577
#3  0x000000000042ac38 in node_foreach_child_full (node=<optimized out>, func=0x42ac80 <node_update_subscription>, params=1, 
    user_data=0x0) at node.c:577
#4  0x00000000004241ef in feedlist_init (fl=<optimized out>) at feedlist.c:221
#5  0x00007ffff14b3329 in g_type_create_instance () from /usr/lib/libgobject-2.0.so.0
#6  0x00007ffff14952bb in ?? () from /usr/lib/libgobject-2.0.so.0
#7  0x00007ffff1496ba1 in g_object_newv () from /usr/lib/libgobject-2.0.so.0
#8  0x00007ffff14974d4 in g_object_new () from /usr/lib/libgobject-2.0.so.0
#9  0x00000000004242da in feedlist_create () at feedlist.c:683
#10 0x00000000004466c5 in liferea_shell_create (app=app@entry=0x70d230, overrideWindowState=<optimized out>) at liferea_shell.c:1355
#11 0x000000000042f037 in on_app_activate (app=0x70d230, user_data=0x0) at main.c:86
#12 0x00007ffff148ff45 in g_closure_invoke () from /usr/lib/libgobject-2.0.so.0
#13 0x00007ffff14a1f91 in ?? () from /usr/lib/libgobject-2.0.so.0
#14 0x00007ffff14aad2c in g_signal_emit_valist () from /usr/lib/libgobject-2.0.so.0
#15 0x00007ffff14ab05f in g_signal_emit () from /usr/lib/libgobject-2.0.so.0
#16 0x00007ffff19af416 in ?? () from /usr/lib/libgio-2.0.so.0
#17 0x00007ffff19af5ea in g_application_run () from /usr/lib/libgio-2.0.so.0
#18 0x0000000000419e43 in main (argc=1, argv=0x7fffffffe338) at main.c:258
```

annoying258:
This is outdated. Feel free to close.

Disable feed auto-updating on start-up and giving useless red marks for all feeds by going to Tools->Preferences,  and uncheck `update all subscriptions at startup`.

lwindolf:
@hugoArregui What do you think about the fix in #730? Could this solve your issue?

Would be great if you can retest git master

hugoArregui:
Sorry @lwindolf I'm not using liferea anymore, I totally forgot about this PR, I'll close it to avoid noise, you can reopen it if you think there is something useful in here. Thanks!

lwindolf:
Thank you for the feedback!

PR #328, opened 2016-03-30
Code cleanup and readd minimize to tray on minimize

asl97:
Clicking the trayicon no longer hide the window

this also fix the lambda TypeError and the liferea started hidden bug.


lwindolf:
@asl97 I cannot reproduce the problem that clicking doesn't hide the window. Can you give a hint? Which desktop environment are you using?

I can reproduce the lambda issue though.

lwindolf:
Ah... got it. I misread minimize as closing. I can reproduce it.

asl97:
probably should have phrase `Clicking the trayicon no longer hide the window` as `Disable minimizing via clicking the trayicon`

Issue #282, opened 2015-10-27
Liferea not start


Drakone:
Hello, liferea not start.
Can you help me ?

	[tom@ordinatom ~]$ liferea --debug-all
	CONF   : proxy auto detect is configured
	CONF   : Manual proxy settings are now NULL:0 NULL:NULL
	NET    : proxy set to http://(null):(null)@(null):0
	DB     : Opening DB file /home/tom/.local/share/liferea/liferea.db...
	DB     : executing SQL: PRAGMA journal_mode=WAL
	DB     :  -> result: 0 (success)
	DB     : executing SQL: PRAGMA page_size=32768
	DB     :  -> result: 0 (success)
	DB     : executing SQL: PRAGMA synchronous=NORMAL
	DB     :  -> result: 0 (success)
	DB     : current DB schema version: 10
	DB     : No VACUUM as freelist count/page count ratio 0,64 <= 10
	DB     : executing SQL: CREATE TABLE items (   item_id		INTEGER,   parent_item_id     INTEGER,   node_id		TEXT,   parent_node_id     TEXT,   title		TEXT,   read		INTEGER,   updated		INTEGER,   popup		INTEGER,   marked		INTEGER,   source		TEXT,   source_id		TEXT,   valid_guid		INTEGER,   description	TEXT,   date		INTEGER,   comment_feed_id	TEXT,   comment            INTEGER,   PRIMARY KEY (item_id));
	DB     :  -> result: 1 (table items already exists)
	DB     : executing SQL: CREATE INDEX items_idx ON items (source_id);
	DB     :  -> result: 1 (index items_idx already exists)
	DB     : executing SQL: CREATE INDEX items_idx2 ON items (comment_feed_id);
	DB     :  -> result: 1 (index items_idx2 already exists)
	DB     : executing SQL: CREATE INDEX items_idx3 ON items (node_id);
	DB     :  -> result: 1 (index items_idx3 already exists)
	DB     : executing SQL: CREATE INDEX items_idx4 ON items (item_id);
	DB     :  -> result: 1 (index items_idx4 already exists)
	DB     : executing SQL: CREATE INDEX items_idx5 ON items (parent_item_id);
	DB     :  -> result: 1 (index items_idx5 already exists)
	DB     : executing SQL: CREATE INDEX items_idx6 ON items (parent_node_id);
	DB     :  -> result: 1 (index items_idx6 already exists)
[...]

PR #344, opened 2016-05-11
Fix startup slowdown caused by huge favicons

mozbugbox:
Huge favicons can cause Liferea startup time increase, up to a few minutes. The favicon was scaled in `favicon_load_from_cache` which could be slow for large(in dimension) pixbuf.

Sites like Yahoo news use .svg file as favicon. An svg can specify random width, height that might be huge. In yahoo's case, the size is 3024x3024. It is really slow to scale such a huge pixbuf. And a user can subscribe to more than a bunch of feeds from that kind of site.

The patch limits the saved size of the favicon file. For already cached huge favicon, user need to remove them from `~/.cache/liferea/favicons`.

I can see issue #282 has some huge performance hit in `import_parse_outline`. Up to 20 seconds for a single call. Don't know if it's caused by huge favicon, though.

### DIFF
--- a/src/favicon.c
+++ b/src/favicon.c
@@ -123,6 +123,21 @@ void favicon_remove_from_cache(const gchar *id) {
 	debug_exit("favicon_remove");
 }
 
+/**
+ * prevent saving huge favicons loaded from net
+ *
+ */
+static void
+favicon_pixbuf_size_prepared_cb (GdkPixbufLoader *loader, gint width, gint height, gpointer user_data)
+{
+	gint max_size = 256;
+	if (width > max_size || height > max_size) {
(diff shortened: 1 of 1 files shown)


lwindolf:
@mozbugbox Thanks for this improvement. I also know some websites with rather huge icons even GIFs.

GreenLunar:
> And a user can subscribe to more than a bunch of feeds from that kind of site.

Regardless this issue, I think Liferea should not store an icon per feed. If user is subscribed to feeds http://lzone.de/taxonomy/weather and http://lzone.de/taxonomy/traffic, there should be a single icon for both under the name `lzone.de.png`.

Would you like me to open an issue for this?

PR #345, opened 2016-05-11
Removes itemSet creation in vfolder_import

Leiaz:
The itemSet is already created in vfolder_new().

I re-read multiple times to make sure. This one shouldn't break anything :)


lwindolf:
Yep. Looks like an unnecessary leak.

PR #335, opened 2016-04-13
Removes the warnings when using combined view

Leiaz:
When itemview_clear is called, item_list_view_clear was called even when the ItemListView widget has been destroyed.
This generated many warnings like : assertion 'GTK_IS_TREE_VIEW (tree_view)' failed ...

PR #347, opened 2016-05-11
Revert "Removes the warnings when using combined view"

lwindolf:
Reverts lwindolf/liferea#335

### DIFF
--- a/src/ui/itemview.c
+++ b/src/ui/itemview.c
@@ -152,9 +152,7 @@ itemview_class_init (ItemViewClass *klass)
 void
 itemview_clear (void) 
 {
-	if (itemview->priv->itemListView) {
-		item_list_view_clear (itemview->priv->itemListView);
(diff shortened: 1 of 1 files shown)


lwindolf:
@Leiaz Reverting this until we find a good solution.

Leiaz:
It would probably be better to first refactor the code a bit. I saw some "TODO: use item merger signal instead". Having the views connect to signals to update would simplify things. 

Currently ItemView relies on ItemListView to check if an item is displayed, even when there is no item list view widget displayed. Also the comment for `item_list_view_create` says `transfer none` but it returns a new ItemListView. 

I could help with that (or other refactoring, I saw you mentioned deriving types from GObject to use introspection for plugins), but I'm not sure where to start.

I can also try to wait patiently :D but if you have some smaller changes to do that would be easy for you to review ...

Issue #349, opened 2016-05-15
Space to skim through articles is not working. (Debian Bug# 706056)


Sidicas:
Space to skim through articles is not working.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=706056

It seems this issue was previously fixed.
https://sourceforge.net/p/liferea/mailman/message/9639687/

However, it is not currently working in the following versions:

Package: liferea
Version: 1.10.12
Version: 1.10.19

Let me know if you need any more info.

lwindolf:
Merged PR from @mozbugbox. Should be fixed now. Please retest!

PR #358, opened 2016-06-26
Fix issue #349: Space to skim through articles

mozbugbox:
Pass skim <space> key only if HTML widget has focus


lwindolf:
Thanks a lot for the fix!!!

PR #360, opened 2016-07-15
Fixes #348: Download content that isn't displayed

### LINKED ISSUE #348: Liferea does not show any errors or prompts when clicking on some types of content. (Debian Bug # 660597)
Liferea does not show any errors or prompts when clicking on some types of content.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=660597

Comments in the bug say that you need to "Register a prompt service"

There is also some example code in the bug report.

Is this something that can be fixed on the liferea side? Or perhaps a message that it juts can't be opened.

Currently, if you click on something that can't be opened, there is no feedback to the user at all.  So the user doesn't understand why their .pdf link doesn't open.

Found in:
Package: liferea
Version: 1.10.12
Version: 1.10.19

Leiaz:
Adds callbacks for the "download-requested" signal, and the
"mime-type-policy-decision-requested" signal to request download of
unsupported MIME types.

I'm using enclosure_download like the "Save Link As" menu item, otherwise we would have to display the status of the download and handle errors ...

### DIFF
--- a/src/webkit/webkit.c
+++ b/src/webkit/webkit.c
@@ -28,6 +28,7 @@
 #include "browser.h"
 #include "conf.h"
 #include "common.h"
+#include "enclosure.h"
 #include "ui/browser_tabs.h"
 #include "ui/liferea_htmlview.h"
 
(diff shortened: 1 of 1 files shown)


lwindolf:
Great enhancement. Wanted to research this for a long time, but never got to it!

PR #361, opened 2016-07-17
Show all enclosures

Leiaz:
I have multiple feeds that show the icon indicating enclosures, but no enclosures are listed because they don't have a audio or video MIME type. It looks like there is no problem opening them with the way Liferea currently associate a command with a MIME type or file extension.

I also did other small changes as separate commits if you want to cherry-pick. I'm using Archlinux and Gtk 3.20 where the "rules-hint" property doesn't do anything, and it's been like that for a while, but it may still be supported by some version of Ubuntu for some time ...

### DIFF
--- a/src/enclosure.c
+++ b/src/enclosure.c
@@ -81,7 +81,7 @@ enclosure_from_string (const gchar *str)
 		return enclosure;
 	}
 	
-	fields = g_regex_split_simple ("^enc:([01]?):([^:]+):(\\d+):(.*)", str, 0, 0);
+	fields = g_regex_split_simple ("^enc:([01]?):([^:]*):(\\d+):(.*)", str, 0, 0);
 	if (6 > g_strv_length (fields)) {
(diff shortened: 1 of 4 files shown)


Leiaz:
Ok now I see the problem with enclosures. The "finding a file extension" part is bugged and will "find" a `.(null)` extension instead of falling back to `data`. I'll try to fix it. It would be nice to have a way to still at least download them. The MIME type in `<media:content>` tags is optional and some feeds don't bother specifying it ...

Leiaz:
The "(null)" was coming from `enclosure_values_to_string`. A NULL mime type was encoded as "(null)" so the MIME type was never NULL. I modified it to leave the field empty, as in the examples, but the this means the items that have been fetched previously have MIME type "(null)".

Also trying to find an extension sometimes result in extensions like `com%2Favatar<long hash>`.

lwindolf:
Thanks for all those fixes/refactoring!!!

PR #382, opened 2016-09-07
Fix proxy settings

Leiaz:
Both proxy settings "Auto Detect" and "No Proxy" result in a NULL
proxy hostname, which result in the default proxy being used at the
start with both options, and no proxy being used after changing to
either option.

This modifies functions to pass the proxy setting mode along with other
proxy settings and to set the SoupSession "proxy-uri" or
"proxy-resolver" properties depending on the setting mode, for both
network and webkit.

Requires libsoup >= 2.42 for the "proxy-resolver" property.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -197,7 +197,7 @@ conf_proxy_reset_settings_cb (GSettings *settings,
 		  proxyusername != NULL ? proxyusername : "NULL",
 		  proxypassword != NULL ? proxypassword : "NULL");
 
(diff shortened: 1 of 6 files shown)


lwindolf:
@Leiaz I merged this for master (1.12-RC1 with Webkit2) with an empty proxy setting due to the missing Webkit2 interface. I plan to merge you webkit2-with-proxy path later with webkit2 version ifdefs.

PR #392, opened 2016-09-18
Make new WebKit related code compile on Debian 8

Leiaz:
Debian 8 has Glib 2.42 and Gtk 3.14.

- Replace Glib 2.44 GType macros by previous style.
- Avoid use of gtk_widget_get_action_group which is new to Gtk 3.16.

@lwindolf I saw you mentioned Debian 8 on the blog post. Now it really works on Debian 8 :)


lwindolf:
Thanks! I'm mostly testing on Ubuntu Xenial and only ~2 weeks compile on Jessie.

PR #393, opened 2016-09-19
Set WebKit font size

Leiaz:
Using Epiphany's code to calculate actual font size in pixels.

The font was just a bit smaller here and it actually looked nice. It looked like a feature not a bug :D 

I also noticed there is a USER_FONT setting key but no option in the GUI to set the font ?

### DIFF
--- a/src/webkit/webkit.c
+++ b/src/webkit/webkit.c
@@ -24,6 +24,7 @@
 
 #include <webkit2/webkit2.h>
 #include <string.h>
+#include <math.h>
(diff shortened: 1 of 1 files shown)


lwindolf:
The USER_FONT (as well as the DOCUMENT_FONT) key is pointing to GNOMEs default font (there is one for document and one for application). So it is exposed in the GNOME settings.

lwindolf:
Works great!

PR #396, opened 2016-09-24
Creates LifereaApplication type

Leiaz:
Currently when Liferea is already started, another call to liferea or
liferea --add-feed will run `db_init`, etc ... But this is only needed
in the primary instance. By using GApplication's "startup" signal, the
init is only done in the primary instance. And the "shutdown" signal
doing the deinit is also handled only in the primary instance.

Using the "handle-local-options" signal from GApplication allows some
options like version to be handled in the local instance. I didn't find
an equilavent to set_summary set_description and set_translation_domain
when adding the option entries to GApplication. The translation still
works.

Opening an url (without the --add-feed option) calls open on the primary
instance without extra code needed.

The feed:https:// scheme is handled in the "open" callback, as GFile
appears to be mangling it a bit ... That way the liferea-add-feed script
is no longer needed, subscriptions can be added by opening urls with
liferea (and the add-feed option is also not needed).

@lwindolf  By the way, thanks for the new powers :)


GreenLunar:
Very good!

paulgevers:
When this is merged, I suggest to enhance the "exec" field in the desktop file as well (I believe it should be "%u").

Leiaz:
We could also add [DBus application launching](https://wiki.gnome.org/HowDoI/DBusApplicationLaunching).

Issue #321, opened 2016-03-10
The date column goes before the title column


birdie-github:
This is a bug unless your native language is Arabic.

Every other Westerner reads from left to right which means your eyes always stumble on the article date which is plain wrong because a feed reader is meant to give you NEW information so date is of the least importance.

If this reasoning is not enough then consider the fact that most other RSS clients show the title column first (on the left).

This is not a feature request. This is a bug report.

lwindolf:
Good idea. We should do this for 1.13+

PR #399, opened 2016-09-29
321 date column

mikelolasagasti:
Fix proposal for #321 

Invert order of columns, setting fixed width for date to 200 and filling space with headline.


Leiaz:
With your change the favicon ends up at the end of the line. I think it looks better in front of the headline.

mikelolasagasti:
Didn't realize about that. Now favicon is before the headlines and the attachment column between headline and date.

Leiaz:
Nice, thanks !

Leiaz:
The click to toggle flagged status (line 661) is broken by the column move.

mikelolasagasti:
@Leiaz thanks for spotting that issue!

elpocko:
I prefer to resize the date column to its minimum width, so it is not visible anymore and does not take up space. I do this because the article date is irrelevant to me.

Would this change break this? Is it still possible to remove the date column?

GreenLunar:
Perhaps a GSettings switch.

mikelolasagasti:
Current status: enclosure and date column are on the left
liferea-5
liferea-6

With this patches: enclosure and date column are moved to the right and date has a fixed size so title has all the remaining space
liferea-4
liferea-3

elpocko:
Ah, thank you. Date is still irrelevant to me, but at least it won't get in the way anymore. That seems like a good compromise.

lwindolf:
@mikelolasagasti  The 200px hard-coded do prevent me from merging this... It should be auto-sizing because 200px mean totally different things on a 1280px wide screen width vs mad HiDPi screen widths...

mikelolasagasti:
@lwindolf I'll explore other options

mikelolasagasti:
@lwindolf now it uses GTK_TREE_VIEW_COLUMN_AUTOSIZE and it seems to do the trick but currently just have my laptop to test.

Leiaz:
Now there are two date columns :D 
It looks like merging went wrong.

The default GTK_TREE_VIEW_COLUMN_GROW_ONLY should work too. Also the doc says :

> Please note that GTK_TREE_VIEW_COLUMN_AUTOSIZE are inefficient for large views, and can make columns appear choppy.

mikelolasagasti:
Thanks @Leiaz ! 

I removed the duplicated date column and changed to use GTK_TREE_VIEW_COLUMN_GROW_ONLY.

I tested with Spanish locale and it resizes correctly in my laptop.

image

Leiaz:
You still have two enclosure icon columns. Don't forget to change the attribute for the icon from "pixbuf" to "gicon".

mikelolasagasti:
@Leiaz thanks again for reviewing. I removed the extra column and fixed the flag toggling logic when the favicon is not present.

Leiaz:
It looks good 👍

lwindolf:
@mikelolasagasti Great work! Having the date out of the way makes focusing on the story titles much more pleasant.

PR #404, opened 2016-10-02
Minor documention fixes

paulgevers:
I spotted some minor glitches in the documentation: one English page contained German text and the version number mentioned on the starting page was still 1.10.


lwindolf:
Thanks!

PR #407, opened 2016-10-07
Replacing deprecated elements in prefs.ui

Leiaz:
I changed target Gtk version to 3.14 (Debian stable version).

GtkHBox, GtkVBox and GtkTable are deprecated, replaced by GtkGrid.
GtkAlignment is deprecated, replaced by setting GtkWidget alignment and
margin properties.

The "margin-left" and "margin-right" GtkWidget properties are deprecated
and should be replaced by "margin-start" and "margin-end" but Glade uses
the old properties https://bugzilla.gnome.org/show_bug.cgi?id=722514 . I
replaced Glade's "margin_left" by "margin_start". The property is not
visible in Glade but it isn't removed when editing the file.

Glade marks the horizontal alignment property of labels as deprecated,
but it corresponds to the "xalign" property, which is a deprecated
property of GtkMisc since 3.14 and a new property of GtkLabel since
3.16. So I left them.

The "xalign" property of (check) buttons used to align the text in their
child widget (label) is deprecated. The doc suggest "access the child
widget directly if you need to control its alignment". I used instead
the "halign" GtkWidget property, which when set to "start" make the
whole button and its text align to the left in its allocated space.

I tried to replace the GtkInfoBar on the Feeds page by a GtkLabel with the "info" style class, as we don't need the action area, but it doesn't look the same. There is a bigger space in blue around the text with the InfoBar.

I wanted to add a message in the same style in the proxy settings panel, that could be hidden when using a future version of WebKit. Something like "Those proxy settings are used only for downloading the feeds. The WebKit widget uses Auto Detect." Is that a good idea, and wording ? It could be confusing because images in articles will be fetched by WebKit.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,7 +29,7 @@ PKG_PROG_PKG_CONFIG()
 ################################################################################
 # Mandatory library dependencies
 
-pkg_modules="	gtk+-3.0 >= 3.4.0
+pkg_modules="	gtk+-3.0 >= 3.14.0
 		glib-2.0 >= 2.28.0
 		gio-2.0 >= 2.26.0
 		pango >= 1.4.0 
--- a/glade/prefs.ui
+++ b/glade/prefs.ui
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.16.1 -->
+<!-- Generated with glade 3.20.0 -->
 <interface>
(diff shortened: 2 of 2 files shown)


lwindolf:
@Leiaz I'm not sure about the proxy message, the topic itself is too complex to explain in one sentence. Maybe linking to a help page explaining details would be a good solution.

Issue #384, opened 2016-09-08
Incorrect display of URL in Source


txtsd:
`Source` links like `https://medium.com/feed/@example_name/` show up as `https://example_name/` in the `Source` field between the list and the article.

lwindolf:
Reproduced. I suspect the password removal code stripping to much.

Leiaz:
This is fixed by #416.

PR #416, opened 2016-10-15
Check URI to check if it has user:password or not

mikelolasagasti:
#384 URI can't contain other than user:password between :// and @ so if it has a / then it has no user/password and print full URI.

It doesn't check if it has user:password and an extra @ char (password protected medium source for example).


lwindolf:
Nice XSLT solution. Surely arcane knowledge :-)

PR #425, opened 2016-10-19
Add GeoRSS info on header

mikelolasagasti:
In https://github.com/lwindolf/liferea/commit/5f92f23ed977759de6e5b0147d039ea92605770b and https://github.com/lwindolf/liferea/commit/19a674be74bfa2e49e30333d95ab7c07bf1af774 GeoRSS map support is removed for security reasons because it depends on an external service. 

This patch adds the coordinates info and a link to OpenStreetMap if the feed has GeoRSS information on it.

### DIFF
--- a/xslt/item.xml.in
+++ b/xslt/item.xml.in
@@ -243,7 +243,29 @@
      </span></b>
    </td>
  </tr>
-</xsl:if>  
+</xsl:if>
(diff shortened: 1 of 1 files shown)


lwindolf:
Nice alternative!!!

PR #429, opened 2016-10-20
Web extension

Leiaz:
@lwindolf I pushed those just after you merged the previous pull request. I noticed when looking for the back button bug that the menu code was still in liferea_htmlview.c.


Leiaz:
I moved those commits on another branch I'm going to push as I modified the same files.

PR #432, opened 2016-10-22
Fixing a navigation/history bug

### LINKED ISSUE #423: 1.12-RC1 Internal browser shows files system on backup
This report is taken straight from the Debian bug 841318¹, as I don't believe this is Debian specific.

¹ https://bugs.debian.org/841318

Became alarmed this morning when using the Liferea internal browser to
view a page from it's feed. When on the webpage, I clicked on an image
included in the page to enlarge it. When finished, I clicked the back
arrow in Liferea viewing area to go back to main page, and was taken
to the root of my file system, allowing me to view each file/directory as a hyperlink.
The desired outcome would be to take me to the original webpage, after
viewing enlarged thumbnail from said webpage.

selection_10-19-1611-08-32001

Leiaz:
This probably fixes #423.
Sometimes the "file:///" url could be added to history ... Details in the commit comment.


lwindolf:
Thank you for fixing this! And also for the header cleanup!!!

PR #437, opened 2016-10-26
Set A-IM in more conditions

Aeyoun:
Change to set the A-IM header if either the If-None-Match or the If-Modified-Since headers are set.

My mistake for only setting it for If-None-Match in [my first implementation](https://github.com/lwindolf/liferea/pull/434). I have later learned that other implementations support both, and that is the most flexible option anyway.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -157,11 +157,17 @@ network_process_request (const updateJobPtr const job)
 		soup_date_free (date);
 	}
 
-	/* Set the If-None-Match: and A-IM: headers */
+	/* Set the If-None-Match header */
 	if (job->request->updateState && update_state_get_etag (job->request->updateState)) {
 		soup_message_headers_append(msg->request_headers,
 					    "If-None-Match",
 					    update_state_get_etag (job->request->updateState));
+	}
+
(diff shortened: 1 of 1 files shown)


Leiaz:
This doesn't compile. You may have accidentally pushed an incomplete commit :  `update_state_get_etag` became `update_state_get_`.

Aeyoun:
Sorry about that, it was getting late and I rushed to fix my mistake. Should be fixed now. :upside_down_face:

Leiaz:
Almost :D. There are too many parentheses.  (`&&` has higher precedence than `||`, clang and Vim + YouCompleteMe show a warning for those `a && b || c` that could be unintended.)

Aeyoun:
> There are too many parentheses.

My clipboard history shows two versions of this code, and I of course committed the wrong one. This should be it; and if not I’ll wait till spring and more sunlight before attempting this again. I do apologize for wasting your time like this.

Leiaz:
No need to apologize. Thanks for the patch !

lwindolf:
Thanks!

PR #439, opened 2016-10-27
Fixing #438 : Ignoring shortcut keys on Web View

Leiaz:
Getting the type name for this feels a bit hacky though ...


lwindolf:
@Leiaz I agree. Still I could think of no other way to safely catch/leave key events.

PR #441, opened 2016-10-30
Update fr.po

Surfoo:


po/fr.po
```diff
@@ -2224,7 +2224,7 @@ msgstr "En cours de téléchargement"
 
 #: ../glade/simple_search.ui.h:9 ../glade/update_monitor.ui.h:4
 msgid "Cancel _All"
-msgstr "_Annuler le tout"
+msgstr "Tout _Annuler"
```
fitojb: I don’t think it’s correct in French to copy English’s “title case” style.

Surfoo: I'm french, yes it's correct :)

fitojb: _Annuler_ is a verb and not a proper noun; a capital letter doesn’t correspond there. See https://fr.wikipedia.org/wiki/Usage_des_majuscules_en_fran%C3%A7ais

Surfoo:
Yes indeed, It's updated!

Issue #29, opened 2014-05-24
Do not use bold text for feeds/folders with unread items in the leftmost treeview


nekohayo:
I just realized something... the use of bold type for "unread items" in the feed list on the left is redundant:
- We already have very nice indicators/"counters" hinting at the exact amount of unread items for each feed/folder
- The bold type takes up more space, allowing less words to fit in and creating many unnecessary ellipses in my case
- It's visually overloading; in most cases, people will have almost every item in bold (especially if using folders to regroup feeds)

I humbly suggest just removing the bold from there and leaving that job to the "counter" widget.

ao2:
FWIW I preferred the previous behavior, it was more consistent with how my email client behaves (Sylpheed), and folders with new content were **popping out** more to my lazy brain.

I guess I'll just have to get used to the new behavior. :)

nekohayo:
AFAIK this didn't change inside the feed view, just on the list on the left, right?
Anyway, my arguments are above and still stand... in my case every item is always in bold as I can't possibly consume all of the content produced by the Internet on a regular basis, and the bold overpowers the "unread counters" visually so it's actually detrimental to attracting my attention; YMMV etc.

ao2:
The change of behavior occurred only in the folder view, yes.

It also depends on **how** you group by folders, of course.
In my case I can notice that the update period for some folders is systematically smaller than other folders, so the bold text make me see when the frequent ones get updated.

Also I have a couple of folders always **expanded** in order to see the individual feeds, and the bold text allows me to pinpoint the updated feeds in a folder more quickly.

But that's just me.

shartge:
I am sorry to pull this old issue back, but Debian just recently upgraded to 1.12-RC1.

This change really reduces the usability of liferea for me. Before, I was able to quickly see which feeds had updated with a simple glance, but now, I have to really look at the light grey square at the right side of the feed name.

You said yourself in 2014: "It is slightly offsetting though that we divert from the email client look" and this is also my opinion. (I don't really care that other feed readers also don't use bold text for the feed list, since I don't compare the look and feel of liferea to other feed readers but to Thunderbird.)

My eye (and mind) is trained to look for **bold** text and I would really like to get that look back, maybe as a configurable option?

GreenLunar:
I like the new appearance.

Perhaps there should a switch to toggle new/old appearance.

shartge:
[...]

PR #445, opened 2016-11-10
Add an option to use bold text for feeds/folders with unread items

ao2:
This allows to have back the old 1.10 look of the feed list, where
feeds/folders with unread items had bold text.

As highlighted in https://github.com/lwindolf/liferea/issues/29 some
users prefer the old look because it suits better their environment,
however liferea does not support it "publicly" anymore, the option is
provided only for backward compatibility.

Users who wanted the old behavior have to set the GSettings value
"feedlist-unread-items-bold" to true by themselves, using dconf-editor
or with a command like the following:

  $ gsettings set net.sf.liferea feedlist-unread-items-bold true

### DIFF
--- a/net.sf.liferea.gschema.xml.in
+++ b/net.sf.liferea.gschema.xml.in
@@ -127,6 +127,11 @@
       <summary>Determines the style of the toolbar buttons</summary>
       <description>Determines the style of the toolbar buttons locally, overriding the GNOME settings. Valid values are "both", "both-horiz", "icons", and "text". If empty or not specified, the GNOME settings are used.</description>
     </key>
+    <key name="feedlist-unread-items-bold" type="b">
+      <default>false</default>
+      <summary>Use bold text for feeds/folders with unread items</summary>
+      <description>Use bold text for the labels of feeds/folders with unread items in the feed list view.</description>
+    </key>
     <key name="trayicon" type="b">
       <default>true</default>
       <summary>Determines if the system tray icon is to be shown</summary>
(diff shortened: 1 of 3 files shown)


shartge:
For what it's worth: I've been running with this patched into the Debian package since October 14 and have not found any problems, glitches or bugs with it.

I personally would be very thankful if this PR could be merged (and hopefully make into the next Debian release on time).

ao2:
Ping

kotnik:
I hope this patch will get into the next release, it significantly improves readability and processing of feeds, especially if you have a lot of feeds in a folder.

ao2:
@lwindolf any comments about this?

Thanks,
   Antonio

carrotIndustries:
\*bump\* I'd really like to see this PR merged, since I assumed the non-bold unread feeds were a bug.

kotnik:
Sad that this didn't make it into `1.12-rc3` release :(

ao2:
Ping.

yaneti:
Tried a plugin approach in pull request #532  that mught be more acceptable..

ao2:
@yaneti thanks, I haven't tested it yet but an external plugin looks acceptable to me too as a solution, if the core patch gets rejected.

Thanks,
   Antonio

lwindolf:
As @yaneti contributed a plugin which works fine, I'm closing this ticket. I hope everyone is ok with it!

ao2:
OK, I tested the plugin and it work really well, thanks @yaneti.

@lwindolf if you could tag an rc4, distributions will update their packages and more users will benefit from the plugin.

Thanks,
   Antonio

lwindolf:
@ao2 I'm aiming for the 1.12.0 actually :-)

PR #446, opened 2016-11-12
Bug with Advanced new subscription dialog

Leiaz:
The id of the dialog in new_subscription.ui didn't match the name of the file and wasn't found. I also replaced the deprecated elements in other subscription ui files.


lwindolf:
Thanks for the great work!!!

PR #450, opened 2016-11-19
Include git compile instructions in INSTALL

da2x:
It almost feels wrong to change an 11-year old file, but I always look in INSTALL first for these kind of instructions.


lwindolf:
Don't worry about the file age :-)

PR #452, opened 2016-11-19
Switch homepage links to HTTPS

da2x:
Avoids redirects.

    grep -lr "http://lzone.de" . | \
        xargs -I % sed -ie 's;http://lzone.de;https://lzone.de;g' %


lwindolf:
Good idea!

mikelolasagasti:
@lwindolf project description is linking to http

liferea-http

PR #456, opened 2016-12-03
Add liferea.css to fix #363

### LINKED ISSUE #363: Missing some spacing above the internal browser's address bar
The internal browser has a navigation/address bar, but the spacing around it is not uniform, it only has some spacing on the bottom:

liferea tabs look

Instead, it should the same 3px spacing ("border") all around the area, on the four sides.

mikelolasagasti:
Created liferea.css style file to fix #363 

Comments are welcome

### DIFF
--- a/glade/Makefile.am
+++ b/glade/Makefile.am
@@ -22,7 +22,8 @@ glade_DATA = about.ui \
 	simple_subscription.ui \
 	theoldreader_source.ui \
 	ttrss_source.ui \
-	update_monitor.ui
+	update_monitor.ui \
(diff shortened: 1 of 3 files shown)


Leiaz:
I'm not sure adding a CSS file is the correct solution. The margin properties of specific widgets can be set in the ui file.

lwindolf:
@Leiaz I agree that solving the issue via .ui would be preferrable. Still providing the CSS hook will solve several unrelated user requests we already had on styling the GUI manually (e.g. on bad background colors from themes).

Issue #442, opened 2016-11-03
Multi language atom feeds


Cj-Malone:
Liferea doesn't support multi language feeds correctly, ignoring the hreflang attribute on links and xml:lang on other elements.

I created an [example feed](https://feeds.betterfeeds.org/examples/multilingual.atom) so you can see [feed title](https://github.com/lwindolf/liferea/blob/master/src/parsers/atom10.c#L704), [entry title](https://github.com/lwindolf/liferea/blob/master/src/parsers/atom10.c#L456) and links just uses the last element the parser encounters but content however adds the [longest](https://github.com/lwindolf/liferea/blob/master/src/item.c#L101).

Cj-Malone:
Sorry, its just the links that are the issue. So [link](https://github.com/lwindolf/liferea/blob/master/src/parsers/atom10.c#L266) needs to take into account ```hreflang```

lwindolf:
@Cj-Malone Can you retest the fixes of #457 (included in 1.12.0)?

Leiaz:
I haven't fixed this, Liferea still uses the last link encountered. `atom10_parse_link` just returns a link, no language information. That link is directly set as the item or feed link in `atom10_parse_entry_link` and `atom10_parse_feed_link`. We would need to check if an alternative link is more appropriate according to locale or display alternative links in some way.

lwindolf:
@Cj-Malone Reopening

@Leiaz thanks for clarifying!

Issue #419, opened 2016-10-17
Image will not be displayed, unless its URL string is complete


GreenLunar:
Assume a feed is located at `https://lzone.de/`

If a link is incomplete (eg. `/liferea/blog/Liferea+Trick+%235%3A+Privacy+with+a+SOCKS+Proxy`), Liferea automatically adds to it `https://lzone.de/`, which results in https://lzone.de/liferea/blog/Liferea+Trick+%235%3A+Privacy+with+a+SOCKS+Proxy

However, when it is so in images (ie. `</img>`), Liferea does not complete it, even though Liferea allows copying the full location of such image.

In order not to promote the following website (not posted yet), and not to spam your mailboxes, I will post example feed, after I done writing this post.

Leiaz:
Which image is not being displayed in that feed ? Currently all items contain a `noimage_thumb.jpg` image that Liferea displays.

GreenLunar:
Yes, that is the image. I am using Liferea 1.11.17.

I have posted a new link to an ATOM feed; please try it.

Leiaz:
The broken images are those with relative urls and a relative path : `dir/image.jpg`. The relative urls with full path (`/dir/image.jpg`) appear to work.

GreenLunar:
Thank you for checking. I have deleted the example link

GreenLunar:
By the way, the URLs with full path are advertisements.

Leiaz:
This was fixed by #457.

PR #457, opened 2016-12-04
Relative url and Atom parser #419

Leiaz:
What is happening with the feed posted by @GreenLunar in #419 is that the subscription "homepage" metadata is NULL, so `node_get_base_url` returns NULL.

Currently the Atom parser will call `subscription_set_homepage` with all "alternate" link elements given (keeping the last as "homepage"). Alternate links can also be links to translations of the feed. [The link element is also not mandatory for feeds](https://tools.ietf.org/html/rfc4287#section-4.1.1), so homepage can end up NULL. I added parsing "self" links. I don't see a way for Atom feeds to indicate which alternate link (if any) is the homepage, so I left "keeping the last one". There is also the handling of links to other languages #442 that could be done here, but I'm not sure what to do with the links.

I modified `node_get_base_url` to return the source if the homepage is NULL. But I'm not sure the homepage should be used as base URL to resolve relative urls. For example, with services like FeedBurner, feeds with a FeedBurner source url have the original website as "homepage" link. According to [RFC3986](https://tools.ietf.org/html/rfc3986#section-5.1) I think it should be `xml:base` if available, then the source url ?

Also, someone who speaks xslt should modify item.xml.in and feed.xml.in to take into account that homepage can be NULL.

### DIFF
--- a/src/node.c
+++ b/src/node.c
@@ -522,8 +522,12 @@ node_get_base_url(nodePtr node)
 {
 	const gchar 	*baseUrl = NULL;
 
-	if (node->subscription)
+	if (node->subscription) {
 		baseUrl = subscription_get_homepage (node->subscription);
+		if (!baseUrl)
+			baseUrl = subscription_get_source (node->subscription);
+	}
+
 
 	/* prevent feed scraping commands to end up as base URI */
 	if (!((baseUrl != NULL) &&
--- a/src/item.c
+++ b/src/item.c
@@ -137,37 +137,17 @@ item_make_link (itemPtr item)
 	if (!src)
 		return NULL;
 
-	/* check for relative link */
-	if (*src == '/') {
+	/* check for absolute URL */
+	if (strstr (src, "://")) {
+		link = g_strdup (src);
+	} else {
 		const gchar * base = item_get_base_url (item);
-		gchar * pos = (gchar *)base;
-		int host_url_size, i;
 
-		/* Check for schema-less (protocol-relative) link */
-		if (*(src+1) == '/') {
-			/* Find first /, end of protocol part in base url */
-			pos = strstr (pos, "/");
-		} else {
-			/* Find the third /, start of link on
-			* site. */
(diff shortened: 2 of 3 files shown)


src/node.c
```diff
@@ -522,8 +522,12 @@ node_get_base_url(nodePtr node)
 {
 	const gchar 	*baseUrl = NULL;
 
-	if (node->subscription)
+	if (node->subscription) {
 		baseUrl = subscription_get_homepage (node->subscription);
+		if (!baseUrl)
+			baseUrl = subscription_get_source (node->subscription);
```
lwindolf: Are you sure it is good to use the subscription source directly?

Because it could be

     http://example.com/feed.atom

and would make a relative link work as

     http://example.com/feed.atom/somelink

instead of

     http://example.com/somelink

Leiaz: Yes, we need the full base URL. And perhaps we shouldn't use the homepage as base URL at all.

In `http://example.com/feed.atom` `example.com` is the authority and `/feed.atom` is the path. 
We need the path part of the base URL to correctly resolve relative links. Relative links can have an absolute path like `/news.html` or a relative path like `news.html`. In the case of a relative path, it must be merged with the base URL path to get the correct link. In your example both resolve to the same thing : `http://example.com/news.html` but if the base URL was `http://example.com/topic/feed.atom` then `/news.html` should resolve to `http://example.com/news.html` and `news.html` should resolve to `http://example.com/topic/news.html`

[Section 5.2 of the RFC3986](https://tools.ietf.org/html/rfc3986#section-5.2) gives the algorithm for resolving relative links. We don't have to implement it ourselves, it is implemented by [xmlBuildURI](http://xmlsoft.org/html/libxml-uri.html#xmlBuildURI). This is why I modified `item_make_link` in item.c to use `common_build_uri` instead of manipulating the string.

lwindolf: You are right, I actually totally missed the leading "/" meaning append to domain.

PR #463, opened 2016-12-14
issue-415 run subscription filter in thread

rich-coe:
Run the update_apply_filter() in a gdk GTask so that the UI continues to update when feeds are updated and feeds have a processing filter specified.


lwindolf:
Great and simple solution!

lwindolf:
@rich-coe One thing for perfect PRs: please always add a ChangeLog entry!

PR #468, opened 2017-01-03
Updated Le Nouvel Observateur feed address

nirgal:


lwindolf:
Thanks for the update!

PR #481, opened 2017-02-04
Set the proxy for WebKit if version supports it

Leiaz:
Also in net.c : set the path of the SoupUri, to make it a valid SoupUri.
Remove a duplicate const, but I realized, the remaining one isn't
useful either, with the typedef this is a const pointer to a non-const
updateJob.

In liferea_htmlview.c : With the proxy set to an unavailable address,
on some items the location changed to file://, so I changed the
condition to avoid adding file:/// to browser history. It feels hacky
though. There might be a better way to handle it, perhaps a custom uri
scheme instead of file:// for our local content ?

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -115,7 +115,7 @@ network_get_proxy_uri (void)
    last modified string.
  */
 void
-network_process_request (const updateJobPtr const job)
+network_process_request (const updateJobPtr job)
 {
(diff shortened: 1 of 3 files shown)


lwindolf:
Kudos for adding proxy support! I agree that it would be better to have a specific schema for internal files.

PR #487, opened 2017-02-27
Update Turkish translation

emintufan:
After 10 years. :)


lwindolf:
Awesome update! Thank you very much!

PR #491, opened 2017-03-12
Change string Preferences -> Properties

### LINKED ISSUE #490: Wrong name for properties in sub context menu
Regarding 1.12rc2:
In the feed list, the "properties" entry in the subscription context menu, is erroneously named "preferences".

Thanks for making this software, couldn't live without it.

mikelolasagasti:
Fix #490 
Introduced in https://github.com/lwindolf/liferea/commit/07717f80e36420528e1dfc2947b856f7e1a996fa


lwindolf:
Thanks for fixing this!

PR #496, opened 2017-03-31
Replace deprecated elements in mainwindow.ui

Leiaz:
And also the deprecated call to gtk_widget_reparent. It is replaced by
gtk_container_remove/add. The widget is temporarily outside any
container, which would destroy it if it isn't owned by something else.
I modified LifereaHtmlView to make it own the reference.

### DIFF
--- a/glade/about.ui
+++ b/glade/about.ui
@@ -1,8 +1,9 @@
-<?xml version="1.0"?>
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.20.0 -->
 <interface>
-  <requires lib="gtk+" version="2.18"/>
-  <!-- interface-naming-policy toplevel-contextual -->
+  <requires lib="gtk+" version="3.14"/>
   <object class="GtkAboutDialog" id="about">
+    <property name="can_focus">False</property>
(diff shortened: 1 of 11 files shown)


Leiaz:
When testing switching views I had noticed that with a folder selected and an item selected, switching between wide and normal view causes `Gtk-CRITICAL **: gtk_tree_store_get_path: assertion 'iter->stamp == priv->stamp' failed`. So I fixed that.

It happened because `itemlist_set_view_mode`  :
- get the selected item 
- calls `itemlist_unload` which calls `itemlist_check_for_deferred_action` which marks as read/removes the selected item from the folder.
- calls `itemview_select_item` to select the now removed item.

And because `gtk_tree_model_get_path` is still called if no iter is found for the item. I fixed this and removed the FIXME because the [sf bug](https://sourceforge.net/p/liferea/bugs/629/) looks like it was due to the old id_to_iter failling to find an iter. Now it shoudn't crash even if the item is not present anymore.

But the selection is lost on switching views, and Liferea will still try to select an item that isn't there.

I added the small fix for #502 to this branch because I'm already modifying liferea_shell.c.

lwindolf:
Great work!

PR #497, opened 2017-04-03
Register custom url scheme

Leiaz:
Register "liferea" url scheme as local and use "liferea://" instead of
"file://" as base url for generated html.

There is the possibility to [register a callback](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebContext.html#webkit-web-context-register-uri-scheme) to handle url requests with that scheme.


lwindolf:
Good idea. Thanks for implementing!

PR #498, opened 2017-04-07
Disable DNT by default

da2x:
See [Internet Explorer 10 default setting controversy](https://en.wikipedia.org/wiki/Do_Not_Track#Internet_Explorer_10_default_setting_controversy).

All browsers and clients disable by default.


lwindolf:
Thanks for working on this! Following the major browsers does make sense.

PR #511, opened 2017-05-14
Fixes #500 Connects browser entry to callback

### LINKED ISSUE #500: Try add copyq as external browser, but get error.
Hello,

I try to add '/usr/bin/copyq add %s' as manual external browser because I want to collect urls for later reading.  "Item->Open in External Browser" shows the message "Failed to execute child process 'mozilla' (No such file or directory)" but the path is correct I checked it. It would be cool if this works :-)

bye

Leiaz:
Glade copy/paste doesn't copy the signals/callbacks part. I missed that one when updating the ui file.

### DIFF
--- a/glade/prefs.ui
+++ b/glade/prefs.ui
@@ -736,16 +736,6 @@
                       </packing>
                     </child>
                     <child>
-                      <object class="GtkComboBox" id="browserpopup">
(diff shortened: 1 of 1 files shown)


lwindolf:
Thanks for fixing it!

PR #514, opened 2017-05-26
Update Indonesian translation

samsulmaarif:


doc/html/topics_id.html
```diff
@@ -0,0 +1,50 @@
...
+<li><a href="concepts_id.html">Konsep Dasar</a></li>
+<li><a href="subscriptions_id.html">Mengelola Langganan</a></li>
+<li><a href="headlines_id.html">Mengelola Berita Utama</a></li>
+<li><a href="folders_id.html">Mengelola Folder</a></li>
+<li><a href="updating_id.html">Memperbarui Langganan</a></li>
+<li><a href="preferences_id.html">Pengaturan Preferensi</a></li>
+<li><a href="searching_id.html">Pencarian Berita Utama</a></li>
+<li><a href="enclosures_id.html">Lampiran/Podcasting</a></li>
+<li><a href="newsbin_id.html">Mengoleksi Item dalam Kotak Sampah Berita</a></li>
+<li><a href="onlineservices_id.html">Berlangganan ke Layanan Daring</a></li>
```
lwindolf: The <a> tags above do reference files not included in this pull request.

It is ok not to translate them, but in this case we need to reference the English ones...

lwindolf:
I've manually merged your translation and default feed list for 1.12.1.

I'd suggest to avoid the other changes as discussed above.

PR #521, opened 2017-06-24
Fix a few compiler warnings

genodeftest:
Fix a few compile-time warnings about
* deprecation (GtkAlignment)
* documentation
* implicit cast

### DIFF
--- a/src/ui/liferea_shell.c
+++ b/src/ui/liferea_shell.c
@@ -557,7 +557,7 @@ on_close (GtkWidget *widget, GdkEvent *event, gpointer user_data)
 {
 	guint signal_id = g_signal_lookup ("delete_event",  GTK_TYPE_WINDOW);
 
-	if (g_signal_has_handler_pending(widget, signal_id, NULL, TRUE))
+	if (g_signal_has_handler_pending(widget, signal_id, (GQuark) 0, TRUE))
 		return FALSE;
 	liferea_shutdown ();
 	return TRUE;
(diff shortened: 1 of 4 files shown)


src/ui/liferea_shell.c
```diff
@@ -908,8 +908,8 @@ static const GtkActionEntry liferea_shell_action_entries[] = {
 	 G_CALLBACK(on_menu_update_all)},
 	{"MarkAllFeedsAsRead", "gtk-apply", N_("Mark All As _Read"), NULL, N_("Marks read every item of every subscription."),
 	 G_CALLBACK(on_menu_allfeedsread)},
-	{"ImportFeedList", "gtk-open", N_("_Import Feed List..."), NULL, N_("Imports an OPML feed list."), G_CALLBACK(on_menu_import)},
-	{"ExportFeedList", "gtk-save-as", N_("_Export Feed List..."), NULL, N_("Exports the feed list as OPML."), G_CALLBACK(on_menu_export)},
+	{"ImportFeedList", "gtk-open", N_("_Import Feed List…"), NULL, N_("Imports an OPML feed list."), G_CALLBACK(on_menu_import)},
+	{"ExportFeedList", "gtk-save-as", N_("_Export Feed List…"), NULL, N_("Exports the feed list as OPML."), G_CALLBACK(on_menu_export)},
```
lwindolf: Cannot merge the change of the dots "..." as it would break the translations :-(

genodeftest: Ok, I've force-pushed a new commit with this change reverted.

Still, I wonder why the translation breaks if changing the application in a way the compiler suggests… Seems like Unicode is still a brand-new technology for tools such as compilers or gettext or whatever is breaking because of the triple-dot.

PR #523, opened 2017-07-02
Improve tray icon behaviour

City-busz:
- Don't show the main window unconditionally. Otherwise the '-w hidden' command line option is ignored.
- Toggle the main window when the user clicks on the tray icon instead of just showing it.
- Deiconify after the window is iconified and hidden. Otherwise the window won't show up on next toggle.

### DIFF
--- a/plugins/trayicon.py
+++ b/plugins/trayicon.py
@@ -113,10 +113,6 @@ class TrayiconPlugin (GObject.Object, Liferea.ShellActivatable):
         self.minimize_to_tray_minimize_handler = self.window.connect("window-state-event",
                                                                      self.window_state_event_cb)
 
-        # show the window if it is hidden when starting liferea
-        self.window.deiconify()
-        self.window.show()
-
         feedlist = self.shell.props.feed_list
         self.feedlist_new_items_cb(feedlist)
         sigid = feedlist.connect("new-items", self.feedlist_new_items_cb)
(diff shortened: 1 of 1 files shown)


lwindolf:
@City-busz Thanks for taking the time to investigate the tray icon logic. I'm sorry that I cannot the patch. The reason being the many many problems we have out there with users migrating/switching to desktop environments without a tray area anymore (mostly this means Ubuntu Unity and GNOME 3).

For this reason the plugin has to be rather aggressive about ensuring that Liferea is accessible (doesn't silently start invisible). The unconditional show on startup is for this reason. So a user without a tray area has the chance to go to the preferences and disable the plugin.

Same with the toggling. Experience shows the state handling with different systray implementations in conjunction with arbitray window managers can make the window state indeterminate. To workaround this we have the trayicon plugin only transition to fully visible window, no matter what the previous state was.

Those decisions are entirely driven from the support perspective!

City-busz:
Please consider to test my change. It won't break anything for users without a system tray. It just fixes some bugs when the user wants to use the tray icon.

lwindolf:
In fact I cannot test you change. Of course I could test on my setup and in 2 or 3 desktop environment combinations. Still ~3 years of trying to control the tray test case complexity gave me the impression that you cannot tame the beast. I see my responsibility as a maintainer to reduce frustrations in downstream bug reports. I can do this by making the tray plugin behave really defensive.

At the same time everyone has the opportunity to modify the bevhaviour by changing the plugin and using a modified version as you implemented it.

I hope you can understand the decision, it's not about not testing your patch, but feeling that it is impossible to support on the complexity of end user setups.

sc0w:
@City-busz thanks

I have tested and it works like a charm here in mate desktop

I don't understand why this usefully PR is closed, it is the expected behavior

paulgevers:
@sc0w did you read what lwindolf wrote? I am not surprised it works like a charm on Mate. Try KDE.

zwickvin (commented):
Works great

Issue #503, opened 2017-04-30
Liferea deanonymize Tor


user1508:
I use liferea with tor for more privacy. I connect liferea to polipo (http-proxy) on localhost and polipo sends the data to tor-socks-proxy. I'm interested in what's going on so I check the established connections with iftop and I discover a behavier of liferea that deanonymize me. Liferea seems to update the feeds through tor but if I click on any item which include a picture or something else liferea create a new direct connection and ignores the proxy-settings. I rockon that the proxy-setting don't pass to the webkit component. **This is very disappointing** because liferea is an fast and useful program!

By the way javascript should be disabled by default and the cookie file should be read only or removed everytime if the user don't want any cookies.

GreenLunar:
> By the way javascript should be disabled by default and the cookie file should be read only or removed everytime if the user don't want any cookies.

### Cookies
Issues: #108 #270 #324
I set file ./.config/liferea/cookies.txt, which is empty, to read-only.

### Javascript
[Why is NoScript configured to allow JavaScript by default in Tor Browser? Isn't that unsafe?](https://www.torproject.org/docs/faq.html.en#TBBJavaScriptEnabled)
Perhaps Javascript should be disabled by default.

Leiaz:
Yes, unless you are using WebkitGTK+ version 2.15.3 or more, the only proxy setting that works as people would expect is "automatic" (system wide settings). 

@lwindolf Perhaps we could grey out other choices and change the setting to automatic for people with older WebKit version, with a warning dialog ? I don't see a case where people would want to download a feed through a specific proxy but download the external elements in items differently.

lwindolf:
@user1508 @GreenLunar what do you guys think about the solution in #526 (included in recent release 1.12.0)?

PR #526, opened 2017-07-17
Warning for unsupported proxy settings

Leiaz:
This is a proposal to "fix" problems like #503. With proxy settings other than auto and older WebKit, Liferea behaves in a way that would be hard to explain to non-tech people.

This shows a dialog on start to warn the user and change proxy settings to "auto" if WebKitGTK+ version doesn't allow setting the proxy (only if the setting isn't "auto" already).

It also shows an orange "warning" infobar and deactivate proxy settings other than "auto".

I pushed a first commit where I was calling `ui_show_info_box` from `liferea_shell_create`. I didn't put it in `conf_proxy_reset_cb` because the main window doesn't exist yet when it is called and `ui_show_info_box` uses it as parent for the dialog. But we can also create a GtkMessageDialog there without a parent so I amended that commit ... In a way it makes sense as it prevents setting the proxy mode to anything other than auto even with external tools, but I'm not sure it is good style to run a dialog here (as oppoed to having some ... signal ? flag ? to tell the mainwindow to show some notice when it is up, perhaps ?).

### DIFF
--- a/glade/prefs.ui
+++ b/glade/prefs.ui
@@ -938,7 +938,7 @@
                   </object>
                   <packing>
                     <property name="left_attach">0</property>
-                    <property name="top_attach">1</property>
+                    <property name="top_attach">2</property>
                   </packing>
                 </child>
                 <child>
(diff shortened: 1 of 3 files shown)


lwindolf:
I think this is a good solution. Thanks for mitigating the issue of #503 

About the dialog handling in conjunction with a parent window: I'm also to shallow a GTK developer to have any opinion :-(

Issue #540, opened 2017-09-17
Keep Headline-Preview Ratio


Euren:
When the window is resized, the various panes do not resize appropriately. That is, when I make the window vertically smaller, the preview pane below rises up but when I make the window larger again, the preview pane stays large. 

I would like to suggest we keep a fixed ratio between the headlines the preview pane. 

It's a bit frustrating to have to resize the separation line every time I resize the window. Especially when using a tiling window manager which resizes.

Version: 1.12-rc2

Somewhat in line to #374

lwindolf:
Please retest with recent releases!

PR #546, opened 2017-10-04
Resize both panes in normal and wide view

Leiaz:
By making both children of the GtkPaned resize, we avoid having one
"squeezed" into a tiny space and not expanding again when the window is
resized.

This fixes or mitigates #540, I'm not sure the ratio is always kept.


lwindolf:
Nice improvement. Thank you!

PR #559, opened 2017-12-02
Fix #447

### LINKED ISSUE #447: 1.12-RC1: liferea doesn't honor "-w hidden" or "--mainwindow-state=iconified" anymore
This is a copy of bug 844262¹ reported in Debian:

++
Dear Maintainer,
to run liferea I use an application autostart entry in Xfce. Command: 'liferea -w hidden'. This used to start liferea with just the tray icon visible. But the current version starts with
visible main window. 

Same effect when launching from command line:

```
wolfgang@pitr:~$ liferea -w hidden
(liferea:8810): Gtk-WARNING **: Failed to get the GNOME session proxy: The name org.gnome.SessionManager is not owned
(liferea:8810): Gtk-WARNING **: Theme parsing error: gtk-widgets.css:1214:18: Not using units is deprecated. Assuming 'px'.
(WebKitWebProcess:8824): Gtk-WARNING **: Theme parsing error: gtk-widgets.css:1214:18: Not using units is deprecated. Assuming 'px'.
```
Using the option --mainwindow-state=iconified instead of hidden also has no effect.

++

After a minor check I can add the following on my own KDE system:
Interesting, on my system, the hidden option is honored when I have the Tray Icon disabled (which means I can't see liferea anywhere). Once I enable the plugin, liferea shows as you describe.

¹ https://bugs.debian.org/844262

Jipok:
https://github.com/lwindolf/liferea/issues/447

### DIFF
--- a/plugins/trayicon.py
+++ b/plugins/trayicon.py
@@ -113,10 +113,6 @@ class TrayiconPlugin (GObject.Object, Liferea.ShellActivatable):
         self.minimize_to_tray_minimize_handler = self.window.connect("window-state-event",
                                                                      self.window_state_event_cb)
 
-        # show the window if it is hidden when starting liferea
-        self.window.deiconify()
-        self.window.show()
-
         feedlist = self.shell.props.feed_list
         self.feedlist_new_items_cb(feedlist)
         sigid = feedlist.connect("new-items", self.feedlist_new_items_cb)


Jipok:
I have not been using rss for two years now :)

lwindolf:
@Jipok Thank you for the feedback!!!

PR #567, opened 2017-12-09
Update fi.po

artnay:


lwindolf:
Thank you for the update!!!

Issue #259, opened 2015-10-06
Feed update setting: Respect server’s cache settings


Aeyoun:
Currently, the user must make a choice for how often a feed should be updated. However, not all feeds are created equal. Some sites update several times per hours and others only once a week or once a month.

I’d like to see a new option where users choose the max age they want to accept or feeds beore rechecking, and otherwise defer to the feeds‘ `Cache-Control: max-age=seconds` HTTP header to see how often it should be updated. I could imagine the UI looking like this:

    [X] Update feeds as often as indicated by the server,
        but at least every [ Interval in days drop-down ].

In an ideal world, users would waste less of their own and the servers’ bandwidth and only frequently pull the feeds that need to be updated often deferring the rest to slower refresh cycles.

Just a request from an admin on a site with a lot of feeds that are updated at different frequency and cache settings but also observe Liferea users pull in every feed every hour regardless of the suggested cache options.

lwindolf:
Closing this feature request.

PR #580, opened 2017-12-31
Fix cache revalidation, source suggested update interval from average of rss2:ttl, syn:update*, and Cache-Control: max-age

da2x:
The Cache-Control: max-age header field contains the number of seconds a response  is considered fresh. This suggests a default update interval similar to syn:* and channel/ttl.

This implementation looks at syn:updatePeriod/updateFrequency, ttl, and maxage (minus any Age header); and uses the average suggested update time from all three time sources (whichever is set). Fixes an issue where syn namespace wasn’t used if updateFrequency appeared above updatePeriod in the feed (as it would in most alphanumerically sorted feeds).

A new 5 minute minimum suggested update interval is enforced.

The definition of syn:updateFrequency:monthly is changed from 31 to 30 days (much more common).

https://tools.ietf.org/html/rfc7234#section-5.2.2.8
https://tools.ietf.org/html/rfc7234#section-4.2.3

Resolves issue #259.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -49,6 +49,9 @@ network_process_callback (SoupSession *session, SoupMessage *msg, gpointer user_
 	updateJobPtr	job = (updateJobPtr)user_data;
 	SoupDate	*last_modified;
 	const gchar	*tmp = NULL;
+	GHashTable	*params;
+	gint		maxage;
+	gint		age;
 
 	job->result->source = soup_uri_to_string (soup_message_get_uri(msg), FALSE);
 	if (SOUP_STATUS_IS_TRANSPORT_ERROR (msg->status_code)) {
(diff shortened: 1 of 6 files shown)


da2x:
The fixes in commits a51bc58  and eddde30 are bugfixes addressing incorrect behavior in master and not just this branch.

@lwindolf, I realize the lack of tests in the codebase slows down integration of this pull request and #581. Do let me know if you have any questions, though.

da2x:
Anything I can do to help move this forward, @lwindolf?

lwindolf:
@da2x Argh... totally missed this PR. I'm so sorry!

I just need to test. For changes in the updating code it usually is best to run using them one week to look out for crashes/corruptions/leaks.

What I'd like to do for such changes is to encapsulate the business logic in a method and do write a unit test for it. The thing is unit testing is actually preventing by this mad C style automake static library building. It almost impossible to link 2 or 3 source files to the Glib test exec without 3 million unresolved symbols...

lwindolf:
Merged this for now. Will check for stability for now.

PR #581, opened 2017-12-31
Fix redirect location updates, and add HTTP 308 (RFC 7538) support

da2x:
New `network_process_redirect_callback` handles updating source locations for HTTP 301 and 308 permanent redirects. There are no practical differences between these two status codes for Liferea.

> This status code is similar to 301 (Moved Permanently) RFC7231, Section 6.4.2, except that it does not allow changing the request method from POST to GET.

https://tools.ietf.org/html/rfc7538#section-3

This replaces pull request #576.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -43,6 +43,28 @@ static gchar	*proxyusername = NULL;
 static gchar	*proxypassword = NULL;
 static int	proxyport = 0;
 
+
+static void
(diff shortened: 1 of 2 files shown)


da2x:
(Sorry for not adding tests. Thanks for merging!)

lwindolf:
Don't worry!

PR #583, opened 2017-12-31
Add aria2 and Transmission (Gtk) to the list of download tools

da2x:
Adds support for megalinks, Magnet URIs, BitTorrent, and other download descriptors. Both aria2 and Transmission are active and maintained projects.


lwindolf:
Good idea. Especially transmission is a widespread tool.

GreenLunar:
Why download managers are hardcoded?

lwindolf:
No special reason. One advantage is that you can later on fix the command for all users. Which for Mozilla and Firefox was necessary roughly 5 times. Without hard-coding you could not have helped everyone downstream.

PR #584, opened 2018-01-01
Clean out dead sites and redirects in OPML files

da2x:
Adds a new Norwegian OPML.

Also, gives everyone a better first-time experience by removing dead websites and relinking all permanent redirects. More sites now default to HTTPS.


lwindolf:
Awesome security work! Thanks wanted to do this for a long time, but never found the time.

da2x:
I thought of automating this, but unfortunately this has to be done manually to confirm that sites haven’t turned into linkfarms or redirect to adult content.

PR #585, opened 2018-01-03
Remove LANG from User-Agent

da2x:
LANG (e.g. en_GB.UTF-8) serves no purpose in the User-Agent string and it unnecessarily leaks information about the user. Improves user privacy somewhat. (Should be included in `Accept-Language` header if anywhere.)


lwindolf:
Thanks for the contribution!

Issue #579, opened 2017-12-30
Date column positioning


flyingOwl:
I just got updated to version 1.12.0 on my Fedora machine. And I really don't like the change that was described here: #321 and done here: #399 . For me the date column is very important so I never wanted it anywhere else.

I have a few news feeds which I rarely read (say: marking items as read) but I sometimes visit the website. So whenever I open that feed I only want to scroll down to an item I have already seen (while visiting the website the previous day e.g.). While scrolling, I heavily depend on the date column. It's easier to remember the time of the last visit than the last headline I've seen. And now the date column is so far away from the headline that it's really unpleasant to keep an eye on it. It's also not easy to see which date belongs to which headline. I know you can change to width of the headline column, but this messes with long titles and would create a lot whitespace to the right.

I also sometimes open (very active) feeds and want to check the time of the last item. Now my eyes have to pan over the whole screen instead of just a few centimeters.

I would propose to either make the columns draggable, so everyone can change the columns on their own. Or at least add a switch to the settings to toggle the date placement (before the title or after the title).

sitaramc:
I second this.  In many of my feeds, I want to wait 2-3 days before I open the link, because the comments are often more interesting/important than the main article (which I would have probably seen elsewhere anyway).

I definitely do not like this change that forcibly moved the date to the far right.  Please make it draggable.

Speaking for myself, even a simple format spec in the preferences that specifies what column goes where would be fine.  So, for example, the default might be "%F %H %A %D", and people like me and flyingOwl could change it to "%F %D %A %H".  You could also allow strftime format specs for the "%D" part.

yaneti:
Technically a plugin approach similar to the bold-unread plugin would work eg:
```
from gi.repository import GObject, Peas, PeasGtk, Gtk, Liferea, Gdk, Pango

class NrDateFirstPlugin (GObject.Object, Liferea.ShellActivatable):
    __gtype_name__ = 'NrDateFirstPlugin'

    shell = GObject.Property (type=Liferea.Shell)

    def do_activate (self):
        normalviewitems = self.shell.lookup ("normalViewItems")
        self.itemlist = normalviewitems.get_child().get_child()
        self.columns = self.itemlist.get_columns()
        self.itemlist.move_column_after(self.columns[4], self.columns[1])


    def do_deactivate (self):
        self.itemlist.move_column_after(self.columns[4], self.columns[3])
```
But not in this simplistic form because that would be very fragile  wrt columns order and view mode changes

[...]

PR #587, opened 2018-01-04
Allow drag and drop reordering and save position of the date column

yaneti:
New configuration option - last-date-column-pos,
updated on each drag and drop reordering of the date column.

an attempt to pacify the unrest in #579

### DIFF
--- a/net.sf.liferea.gschema.xml.in
+++ b/net.sf.liferea.gschema.xml.in
@@ -232,6 +232,11 @@
       <summary>Send "Do Not Track" header</summary>
       <description>Configures wether the "DNT" header is to be sent. If enabled sends "DNT: 1", meaning Do Not Track.</description>
     </key>
+    <key name="last-date-column-pos" type="i">
+      <default>-1</default>
+      <summary>Date column position</summary>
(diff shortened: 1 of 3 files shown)


lwindolf:
Awesome change! I'll promise to review it soon.

One thing that worries me is that only the date column position is persisted. I already see users putting the state column to the far right and expecting it to stay there...

yaneti:
With the current patch the only column marked reorderable (i.e the only one you can drag/drop) is the date column, so technically the state column can't be moved to the far right.
It is more like an Easter egg than a fully customizable order.

I guess It could be redone with say a  "item_list_colmns" setting holding some variant, hash or string encoding of the order if thats the preferable thing..

yaneti:
Another approach to the column ordering submitted in #589 . Moves the column order in a setting, which covers all currently used columns.

lwindolf:
With #580 being merged what would be the next steps for DnD?

yaneti:
It took some time for me to remember were we were with this ;)

Perhaps the next step is the patch in #629 which actually marks all columns reorderable and saves the new order in the new setting

lwindolf:
Sorry about the delay and thank you for your patience!

PR #596, opened 2018-01-26
Update sv.po

jony0008:
Update Swedish translation


lwindolf:
Thank you for the update!!!

Issue #578, opened 2017-12-30
Unable to set unread items in bold


EverEve:
Hello, 
I use Liferea on Debian Testing since long time. I've started it today and I've found that unread items are not displayed in bold and article date/time is no more displayed.

This happens with self-compiled package and the Debian repository package as well.
Tried also to build from the latest tarball source but the issue is not solved.

Can someone help? Thanks.

lwindolf:
Sounds very strange. I cannot reproduce this. Can you provide a screenshot and check if there are any command line errors while accessing the headlines?

sicelo:
As someone else experiencing the problem, (Debian Sid), I opened a bug against liferea, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=885868.

Without hijacking OPs thread, I am attaching my command line log, as well as a screenshot (you notice there are 5 unread items for the active feed, but on the headline everything looks the same). As mentioned on the Debian bug report, I removed everything in ~/.config/liferea, so hopefully it is not a case of corrupted settings, etc.

[liferea.log](https://github.com/lwindolf/liferea/files/1604030/liferea.log)
![screenshot from 2018-01-04 18-52-07](https://user-images.githubusercontent.com/3158319/34574735-ffd5450c-f180-11e7-82b4-8d6d24ceb5a5.png)

petep:
I too am having the same problem. Here's my setup:

OS: Fedora 27, 32 bit
Version: liferea-1.12.0-2.fc27.i686
Desktop: Mate 1.19.2

Please let me know if you need any further details.

lwindolf:
@sicelo @petep Which views are you using? Email or wide view. Cannot tell from the screenshot...

sicelo:
my screenshot was taken with "Normal" view active.
It looks like the problem does not happen on Wide View (which I do not use, because it is too wide for my 1024x768 screen)

So culprit is "Normal" view.

petep:
I can confirm that. Wide view highlights new articles correctly.

petep:
Corresponding Redhat bug: https://bugzilla.redhat.com/show_bug.cgi?id=1534643

Seems it's tricky to reproduce.

lwindolf:
I do have no idea yet what is causing the issue. But I believe the difference in behaviour between wide and normal view comes from the two different code paths:

- normal view: setting PANGO_WEIGHT_NORMAL or BOLD via gtk_tree_store_set()
[...]

PR #601, opened 2018-02-02
Changing item and feed time from time_t to gint64

Leiaz:
I'm using `gint64` and not `guint64` as I said in #578 because it looks like it is more commonly used for time, as it allows representing dates before epoch too (even if *we* don't need that).

I haven't modified `date.c` for now, but I thought about rewriting it to only use cross platform GLib functions if possible. That way we wouldn't have to deal with the Linux vs Windows vs Mac problems ourselves, and we could remove the macros and conditional compile.


Leiaz:
So, Glib doesn't have a direct equivalent to `strptime` (which is not available in Windows). It has a currently incomplete `g_time_val_from_iso8601` to parse iso8601 dates that will be more complete in Glib 2.56. To avoid using `strptime` we have to parse some dates ourselves, so I'm not sure this is simpler ...

It currently fails the parse date tests for iso8601_hours and iso8601_wrong : Glib's `g_time_val_from_iso8601` reports success in parsing them, and returns a value ...

Leiaz:
@lwindolf Do I add manual parsing of iso8601 to fix the test (is the hour only format actually allowed in feeds ? for the [atom format](https://tools.ietf.org/html/rfc4287#section-3.3) a date must be in the date-time form from [RFC3339](https://tools.ietf.org/html/rfc3339) which includes the full time) ? 

Or do I leave it like this ? Or scratch all of this and just fix the gint64/time_t thing ?

lwindolf:
I would go for adjusting the test. I'm not sure where I took the test example. I'd say I mistakingly thought up a ISO date without the minutes.

lwindolf:
I'm fine with your code. I personally prefer goto's if the simplify the code.

PR #604, opened 2018-02-12
Correctly print error message when failing to unlock gnome keyring

john-smith7:


lwindolf:
Thank you for improving this!

PR #608, opened 2018-02-25
Switch to GAction and GtkBuilder

Leiaz:
This is cherry-picking of commits from #276 and it is not finished. I haven't modified the functions that enable/disable actions yet. I want to see if the actions can be grouped first. Except for that, it should work.

@GreenLunar Are the toolbar buttons in the order you expect for RTL ?

### DIFF
--- a/glade/Makefile.am
+++ b/glade/Makefile.am
@@ -6,6 +6,7 @@ glade_DATA = about.ui \
 	enclosure_handler.ui \
 	google_source.ui \
 	inoreader_source.ui \
+	liferea_menu.ui \
 	mainwindow.ui \
 	new_folder.ui \
 	new_newsbin.ui \
--- a/glade/Makefile.am
+++ b/glade/Makefile.am
@@ -7,6 +7,7 @@ glade_DATA = about.ui \
 	google_source.ui \
 	inoreader_source.ui \
 	liferea_menu.ui \
+	liferea_toolbar.ui \
 	mainwindow.ui \
 	new_folder.ui \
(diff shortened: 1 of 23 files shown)


Leiaz:
The menu item from the plugin isn't removed when the plugin is deactivated. It looks like the only way to remove an item from a menu is [by index](https://developer.gnome.org/gio/stable/GMenu.html#g-menu-remove), and I'm not seeing a nice way to find an item's index and the plugin has no way to know if it was the last to add an item ...

Also, I'll look into how accelerators work now. The [trick to edit them](https://docs.xfce.org/faq) doesn't work.

Leiaz:
Accelerators now work and could be customized from a Python plugin using `gtk_application_set_accels_for_action`.

Leiaz:
@lwindolf  I looked in ui/popup_menu.c to see if we could use the actions there and simplify the code. We could because currently `ui_popup_item_menu` doesn't take the passed item into account, so the popup only works because a right click changes the selected item. Maybe we should fix that ? Sometimes I'd like to right-click open an item without unselecting the one I'm looking at. That is how the treeviews usually work.

lwindolf:
Makes sense. Would be a nice improvement.

Leiaz:
@lwindolf I think I'll stop adding to that branch for now :) If you want to review and merge to master or to your headerbar plugin branch to add some actions to the buttons.

lwindolf:
@Leiaz So far works like a charm. Code looks nice and improved too. I really value your work!

PR #610, opened 2018-03-02
appdata: Set the component type to desktop-application

yaneti:
Appstream spec says type must be set for desktop applications components
https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html

Fixes liferea showing up in current gnome-software


lwindolf:
Thank you for fixing this!!!

PR #613, opened 2018-03-06
Fixes #602 CSS stylesheet not picking link colors

### LINKED ISSUE #602: CSS stylesheet not picking #GTK-COLOR-NORMAL-LINK from GTK theme
I'm using liferea with gnome 3.26, adapta-gtk-theme and Wide view.

I found that the article view does not follow my gtk theme (I only noticed the link colors), and I don't know if it is only in the article view, but here is a screenshot 
omg

another UI thing is favicons, most people use 8x8 or 16x16 favicons for their website, but liferea scales them to fit the area and that pixelates the images
list
I suggest viewing an image from the article not the website favicon is a good idea but I don't know if it is possible, but I have seen other feed readers doing that.

another feature I miss is mark as read by scroll over news items.

thank you for your great work!

lwindolf:
Directly extracting link-color and visited-link-color from GtkStyle seems to have been broken since a while.

Partially migrates color extraction to GtkStyleContext. Fully migrating doesn't seem to be possible as light and dark styles are not exposed by GtkStyleContext as far as I understand it. Might become a problem in GTK4

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,9 @@ To be released
 	* #495: New command line option --disable-plugins (-p) to start
 	  with all plugins disabled.
 
+	* Fixes #602: CSS style for GTK link colors not used
+	  (reported by pupyc)
+
 	* #594: Update of Polish default feed list
 	  (wmyrda)
 	* #584: Fixes broken OPML feed list entries
--- a/src/render.c
+++ b/src/render.c
@@ -1,7 +1,7 @@
 /**
  * @file render.c  generic GTK theme and XSLT rendering handling
  * 
- * Copyright (C) 2006-2013 Lars Windolf <lars.windolf@gmx.de>
+ * Copyright (C) 2006-2018 Lars Windolf <lars.windolf@gmx.de>
  *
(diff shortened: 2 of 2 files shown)


Leiaz:
It looks like the unvisited link color is not used, it is still dark blue when it should be a lighter blue : 
```
HTML   : theme color "GTK-COLOR-LINK" is 4A90D9
HTML   : theme color "GTK-COLOR-VISITED-LINK" is EEEEEC

HTML   : theme color "GTK-COLOR-LINK" is 5588EE
HTML   : theme color "GTK-COLOR-VISITED-LINK" is 2A76C6
```
The visited link color is displayed correctly.

And yes, this will need to be modified when switching to Gtk 4 as even [GtkStyleContext won't work like this anymore](https://developer.gnome.org/gtk4/unstable/gtk-migrating-3-to-4.html#id-1.6.4.3.8).

lwindolf:
Found a typo "GTK-COLOR-LINK" in the code and "GTK-COLOR-VISITED-LINK" in the CSS.

PR #614, opened 2018-03-07
Headerbar support via plugin (discussion in #276)

lwindolf:


plugins/headerbar.plugin
```diff
@@ -0,0 +1,12 @@
+[Plugin]
+Module=headerbar
+Loader=python3
+IAge=2
+Name=Header Bar
+Version=0.1
+Description=Get a GNOME style header bar for Liferea
+Authors=Mozbugbox <mozbugbox@yahoo.com.au>
```
Leiaz: Mozbugbox ? :)

plugins/headerbar.py
```diff
@@ -0,0 +1,90 @@
...
+        ui_manager.get_widget("/MainwindowMenubar/").set_visible(False)
+        self.hb.show_all()
+
+    def do_deactivate (self):
+        ui_manager = self.shell.get_property("ui-manager")
+        ui_manager.get_widget("/maintoolbar/").set_visible(True)
+        ui_manager.get_widget("/MainwindowMenubar/").set_visible(True)
+        self.hb = None
+
+    def _on_back(self, button):
```
Leiaz: This won't be necessary. Buttons are [GtkActionables](https://developer.gnome.org/gtk3/stable/GtkActionable.html) and can be directly associated with an action.

plugins/headerbar.py
```diff
@@ -0,0 +1,90 @@
...
+        button = Gtk.Button()
+        icon = Gio.ThemedIcon(name="edit-find-symbolic")
+        image = Gtk.Image.new_from_gicon(icon, Gtk.IconSize.BUTTON)
+        # FIXME: signal
+        button.add(image)
+        self.hb.pack_end(button)
+
+        self.shell.lookup("mainwindow").set_titlebar(self.hb)
+
+        ui_manager = self.shell.get_property("ui-manager")
```
Leiaz: I changed the property name to "builder" in #608. (And maybe it doesn't need to be a property,  as `liferea_shell_lookup` could be used).

plugins/headerbar.py
```diff
@@ -0,0 +1,90 @@
...
+        ui_manager.get_widget("/maintoolbar/").set_visible(False)
+        ui_manager.get_widget("/MainwindowMenubar/").set_visible(False)
+        self.hb.show_all()
+
+    def do_deactivate (self):
+        ui_manager = self.shell.get_property("ui-manager")
+        ui_manager.get_widget("/maintoolbar/").set_visible(True)
+        ui_manager.get_widget("/MainwindowMenubar/").set_visible(True)
+        self.hb = None
```
Leiaz: The window likely takes ownership of the bar, so this won't remove it. Setting the window's titlebar to None should work.

plugins/headerbar.plugin
lwindolf: Copy&Paste :-)

plugins/headerbar.py
lwindolf: I agree about not using ui_manager. Will rebase it once #608 is merged.

liferea_shell_lookup() cannot be used though (with GtkUIManager at least) as the widget names are not exposed.

plugins/headerbar.py
lwindolf: Well it works. Setting to None might be problematic when reactivating.

Does the titlebar stay visible for you? If yes, which WM do you use?

Leiaz: Yes, the GtkHeaderBar stays. I use XMonad, but the same thing happens in a Debian 8 VM with Gnome.

Why would it be problematic ? A new GtkHeaderBar is created and set as titlebar in `do_activate`.

plugins/headerbar.py
lwindolf: Fixed now.

plugins/headerbar.py
lwindolf: Also resolved.

plugins/headerbar.py
lwindolf: You are right. I changed the deactivate() behaviour to

    self.shell.lookup("mainwindow").set_titlebar(None)

lwindolf:
So far I'd say the plugin is somewhat complete.

Leiaz:
Yes, looks good. :+1:

PR #615, opened 2018-03-07
Fix date_format function prototype

yaneti:
Otherwise biuld fails on i686 and armv7hl


Leiaz:
Whoops, I missed that. Thank you !

paulgevers:
I confirm that with this change 1.2.2 builds on Debian i386.

PR #621, opened 2018-03-31
Building with Meson

Leiaz:
I just wanted to try this Meson everyone is talking about on an actual application. This is not really a pull request, I just wanted to share it, if anyone want to try. It can't coexist with autotools as written. I modified some files to use only gettext and no longer use intltool (maybe we should [do that with autotools too](https://wiki.gnome.org/MigratingFromIntltoolToGettext)), and I moved the icons to just install the directory, not rewrite the path for each icon.

I like it. It doesn't vomit files all over the source tree. The build files are very readable. It automatically generates a `compile_commands.json`, which is used by tools like YouCompleteMe vim plugin. It doesn't have separate instructions for `dist` : it packages the latest commit in version control (compiling and testing it first).


mikelolasagasti:
@Leiaz do you plan to work on this PR to make it mergeable again? If not, I can continue the work

Leiaz:
@mikelolasagasti No I didn't really plan to, as I don't see the point unless @lwindolf is interested in merging it. I'm not using it, because when I build Liferea to test changes, I need to do it the way it is going to be released anyway.
So yes, you can do it, if you are interested.

lwindolf:
@Leiaz @mikelolasagasti I'm not against Meson if the overhead (in declaration and release commands) is less then with autotools I'd be happy with it. We should also think of the downstream maintainers whether Meson works good for them too (I have no clue how useful it is for them).

In general I personally think autotools has seen so little improvement over time, it is time to give some other project a chance.

Leiaz:
@lwindolf  It is the build system used by multiple GNOME projects, like gtk, evince, Epiphany ... I think that is where I read about it. So packagers shouldn't have a problem adapting their scripts. Maybe we could wait for a big version change, so it will be less of a surprise ?

lwindolf:
I wouldn't wait as 1.13.x is defined (in the release tags) as unstable so breaking stuff it ok per definition.

mikelolasagasti:
Meson builds are faster (tested with cached sudo):

```
(master)
/usr/bin/time  -p sh -c 'meson _build/  ; ninja -C _build/ ; sudo ninja -C _build/ install'
 
real 11.06
user 49.54
sys 7.70

(13.3)
/usr/bin/time  -p sh -c './autogen.sh ; ./configure ; make -j9 ; sudo make install'
real 17.06
user 56.73
sys 9.33
``` 

From maintainers perspective, a project I work on migrated to Meson and Fedora, Debian, openSUSE and Arch created new packages/recipes same day, so I don't think it would be a problem for downstream packaging.

I've updated this PR locally and I need to check if i18n works fine before creating a new PR.

lwindolf:
@Leiaz I'm closing this as the new "gtk4" branch is now on Meson.

PR #626, opened 2018-04-11
Avoid icons in main menu

lwindolf:
Since refactoring to GtkBuilder and GAction the
main menu suddenly has menu icons again. This is
inconsistent with all other context menus and doesn't
follow HIG. Also the menu entries with icons are
not aligned with those without icons.

I'd prefer the icons to be hidden by GNOME defaults
so other desktop environments can use them, but I have
no idea on how to achieve it...

### DIFF
--- a/glade/liferea_menu.ui
+++ b/glade/liferea_menu.ui
@@ -8,19 +8,16 @@
         <item>
           <attribute name="action">app.update-all</attribute>
           <attribute name="label" translatable="yes">Update _All</attribute>
-          <attribute name="icon">view-refresh</attribute>
         </item>
         <item>
           <attribute name="action">app.mark-all-feeds-read</attribute>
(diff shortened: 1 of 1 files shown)


lwindolf:
LOL GNOME is as alway a really pain. First they hide all icons because they don't like them. So it works for other desktops. Then they deprecate GtkMenuImageItem in favor of GMenu in which they reintroduce mandatory icons that can not be disabled the same way (last section in https://wiki.gnome.org/HowDoI/GMenu). And of course with ne arbitrary UX rules: icons only for nouns, never for verbs.

As always there is no way to way to win as a non-core application.

According to those rules the icon for menu entry 'Preferences' could stay.

For now I'd say let's drop all and see how the non-GNOME user will react. I fear some of them will be rightfully be mad about it...

PR #629, opened 2018-04-12
Allow dnd item list view column reorder

yaneti:
So as continuation of #587 and #589  allow column reordering by drag and drop.
Save the order in the setting introduced in #589 

The are two commits here. The first one is the reorder patch.
The second one is a cleanup of liferea shutdown on window close button, someone that can use the trayicon plugin should probably check that it does what it means

### DIFF
--- a/src/ui/item_list_view.c
+++ b/src/ui/item_list_view.c
@@ -732,7 +732,6 @@ on_item_list_row_activated (GtkTreeView *treeview,
 static void
 on_item_list_view_columns_changed (GtkTreeView *treeview, ItemListView *ilv)
 {
-	gint n = gtk_tree_view_get_n_columns(treeview);
 	gint i = 0;
 	GList *columns;
 	GHashTableIter iter;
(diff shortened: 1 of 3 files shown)


Leiaz:
You forgot to free the GHashTable.

src/ui/liferea_shell.c
```diff
@@ -569,10 +569,6 @@ on_notebook_scroll_event_null_cb (GtkWidget *widget, GdkEventScroll *event)
 static gboolean
 on_close (GtkWidget *widget, GdkEvent *event, gpointer user_data)
 {
-	guint signal_id = g_signal_lookup ("delete_event",  GTK_TYPE_WINDOW);
-
-	if (g_signal_has_handler_pending(widget, signal_id, (GQuark) 0, TRUE))
-		return FALSE;
```
Leiaz: This does change the way the plugin works. Liferea shuts down instead of minimizing on close.

src/ui/item_list_view.c
```diff
@@ -724,6 +727,37 @@ on_item_list_row_activated (GtkTreeView *treeview,
...
+	   should not be hooked up with less than the full number of columns
+	   eg: on item_list_view creation or teardown */
+	g_return_if_fail (n == 5);
+
+	columns = gtk_tree_view_get_columns (treeview);
+	for (GList *li = columns; li; li = li->next) {
+		g_hash_table_iter_init (&iter, ilv->priv->columns);
+		while (g_hash_table_iter_next (&iter, &colname, &colptr)) {
+			if (li->data == colptr) {
+				strv[i++] = g_strdup (colname);
```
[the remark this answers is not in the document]
yaneti: Actually I think I don't need to dup the strings here. The hastable keys are staic and never disposed of so can be used here directly.

yaneti:
Thanks for the review @Leiaz 

About the plugin. I'll try to dig some more. The signal should be stooped afaikt but maybe there is some pygtk oddity oddity`here
`

src/ui/liferea_shell.c
lwindolf: @yaneti What was the intention of this removal? This is the last thing preventing the merge as I see it.

yaneti: @lwindolf the idea is to properly shutdown on window close button, regardless of the tray plugin state.  The current on_close code never does because on_close itself is still pending. 

I think I properly fixed the plugin after this removal in c78c0dd8abbbc9e712418cc880cbc42fc9e484f7  by blocking the main delete-event handler when the plugin is active.

lwindolf:
@yaneti Works perfectly. The only issue is that on shutdown the safety check in on_item_list_view_columns_changed() triggers an causes GLib-GObject-CRITICALs

    g_return_if_fail (gtk_tree_view_get_n_columns(treeview) == 5);

Several distros core dump on criticals. Therefore I'd suggest to change it to a simple return.

Issue #624, opened 2018-04-07
include some download methods


motherload:
I suggest including youtube-dl and wget as available download methods for enclosures.

lwindolf:
This becomes possible using the referenced change with 1.12.3. Closing this.

PR #634, opened 2018-04-30
Adds settings for custom download command, Fixes #575

### LINKED ISSUE #575: Allow additional enclosure download tools
Problem in short
----------------
The supported *download* applications are hardcoded and limited.
This results is inpossible extensibility without „hijacking“ on of the hardcoded command.

My use case
-----------
I want to download enclosures with a custom and special postprocessing done by a script I written for myself.
Let for the moment reference this script as *myDownloadScript*.

Currently I'm force to add *myDownloadScript* as a link in with a name like *kget* or *uget-gtk* to a location
recognized by my `PATH` statement.

This is an awfull hack → 😱

Surely the hack can be encapsulated in a custom wrapper script for *liferea* – but is still an awfull hack.

Proposal
--------

In *./src/ui/preferences_dialog.c* I see the following code:

```
/** download tool commands need to take an URI as %s */
static const gchar * enclosure_download_commands[] = {
"steadyflow add %s",
"dbus-send --session --print-reply --dest=org.gnome.gwget.ApplicationService /org/gnome/gwget/Gwget org.gnome.gwget.Application.OpenU
RI string:%s uint32:0",
"kget %s",
"uget-gtk %s"
};

/** order must match enclosure_download_commands[] */
static const gchar *enclosure_download_tool_options[] = { "steadyflow", "gwget", "kget", "uget", NULL };
```

My proposal is to add an additional method:
```liferea-user-script %s```

This would allow to install *myDownloadScript* under a name, which would not hide a tool used in other places.
Also it introduce an extension point.
If the script *liferea-user-script* is not existing on the local system it behaves like e.g. for *steadyflow* also
not available on many installation.

Leiaz:
This adds radio buttons and an entry for a custom download command. I didn't put a file explorer because it is simpler for us and allows more complex commands if users add the `%s` for the URL anywhere in the command (like for the custom browser command).

I also removed the unused liststores from the ui file (new ones are created and filled in prefs.c).

This should solve issues #575 and #624.

### DIFF
--- a/glade/prefs.ui
+++ b/glade/prefs.ui
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.20.0 -->
+<!-- Generated with glade 3.20.4 -->
 <interface>
   <requires lib="gtk+" version="3.14"/>
   <object class="GtkAdjustment" id="adjustment2">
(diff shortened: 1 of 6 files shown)


lwindolf:
Several users will really like this!

PR #648, opened 2018-05-29
Fixing unread item count for remote sources, #557

### LINKED ISSUE #557: Unread items count not updated when items are read by another TT RSS client
When using Tiny Tiny RSS items can be read by another client besides Liferea (for example in web interface).  If so Liferea does not update unread items count.

Let's say Liferea shows 2 unread items for a feed Foo in Feed List.  Then these two items are read in the Tiny Tiny RSS web interface.  Going back to Liferea it still shows 2 unread items for Foo.  Refresh nor Mark All As Read does not help.

When Liferea is in this inconsistent state Ctrl + N (Next Unread Item) does not work properly.  It wants to jump to a non existing unread item and refuses to go further to the proper (actually existing) unread items.

This is for version 1.12-rc3.

Leiaz:
When merging received items with the existing list, if the item state
has changed in the remote source, the new state is directly assigned to
the existing item.

This replaces the direct assignment with calls to
`item_read_state_changed`, `item_flag_state_changed` which performs
related operations like updating the unread count.

This should fix #557.

### DIFF
--- a/src/itemset.c
+++ b/src/itemset.c
@@ -29,6 +29,7 @@
 #include "feed.h"
 #include "itemlist.h"
 #include "itemset.h"
+#include "item_state.h"
 #include "metadata.h"
 #include "node.h"
 #include "rule.h"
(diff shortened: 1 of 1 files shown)


lwindolf:
While the change is correct from a functional perspective, I'm a bit worried about performance.

For example. I'm using tt-rss and read ~100 feeds via tt-rss about every 2 days. On each startup I have to sync ~2000-3000 items and the change raises the cost from some in-memory writes to one DB action write + duplicate query + GtkTreeView update for each.

PR #649, opened 2018-05-30
Switch from persistent to session-only cookies

da2x:
Persistent cookies serve no purpose in a feed reader other than enabling user tracking and profiling. This patch switches to session-only cookies; allowing for cookies to remain for the duration of the session. This ensures privacy for users while still retaining compatibility with feeds that go into tamper-tantrum redirect loops if they aren’t allowed to set cookies.


lwindolf:
Setting the file to readonly is a good idea!

I fear though we cannot change the filename. This would break the use case served here for existing setups.

da2x:
The filename must be changed or exisisting cookies stored in the file would never expire. I could have handled upgrade and deleted the old file, but this really wasn't worth the work.

Are you aware of any feeds on the planet that require persistent cookies? All historical data I can find on the web only mention cookie-requirements for avoiding redirect loops from misconfigured servers (feeds from New York Times around 2006, specifically). Session cookies handles this usecase perfectly and discards the cookies when they're no longer needed.

lwindolf:
@da2x I trust your judgement and testing. Change is merged.

PR #652, opened 2018-06-06
Fixes #557: Updating counters for remote sources

### LINKED ISSUE #557: Unread items count not updated when items are read by another TT RSS client

Leiaz:
This one should update the count only once per node.

Also, owned by Microsoft :nauseated_face: :(

### DIFF
--- a/src/itemset.c
+++ b/src/itemset.c
@@ -202,7 +202,6 @@ static gboolean
 itemset_merge_item (itemSetPtr itemSet, GList *items, itemPtr item, gint maxChecks, gboolean allowUpdates)
 {
 	gboolean	allowStateChanges = FALSE;
-	gboolean	html5_enabled;
 	gboolean	merge;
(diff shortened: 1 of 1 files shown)


lwindolf:
Thanks for improving this!

And about MS, let's see live here, how quickly Github will deteriorate, or maybe not.

PR #662, opened 2018-08-05
 Move from GnomeKeyring to libsecret 

bgermann:


plugins/libsecret.plugin
```diff
@@ -0,0 +1,18 @@
...
+IAge=2
+Name=Libsecret Support
+Name[de]=Unterstützung für Libsecret
+Name[he]=תמיכה עבור Libsecret
+Name[it]=Supporto al portachiavi di Libsecret
+Description=Allow Liferea to use libsecret as password store
+Description[de]=Erlaubt Liferea Passwörter mit libsecret zu speichern
+Description[he]=אפשר ליישום Liferea להשתמש בתוכנה libsecret כמחסן סיסמאות
+Description[it]=Consente all'applicazione di usare come archivio delle password il portachiavi di libsecret.
+Authors=Lars Windolf <lars.windolf@gmx.de>
```
lwindolf: Author is wrong :-)

lwindolf:
The new plugin works like a charm. One open problem I see: how to migrate all users to the new plugin?

bgermann:
I thought I tested reading already-stored credentials. Doesn't it?

bgermann:
Or do you mean the plugin registry file? Just leave out the second commit and it is still named gnome-keyring.

lwindolf:
If you don't mind I'd do exactly this and leave the old name so we don't need to do any migration.

bgermann:
Fine.

plugins/gnome-keyring.py
```diff
@@ -23,44 +23,47 @@
...
-            print("There doesn't seem to be a keyring yet.")
+    def get_collection(self):
+        service = Secret.Service.get_sync(Secret.ServiceFlags.LOAD_COLLECTIONS, None)
+        colls = service.get_collections()
+        for c in colls:
+            if c.get_label() == LABEL:
+                return c
+        return None
 
     def do_deactivate(self):
```
Leiaz: Maybe this could be removed. It doesn't do anything.

bgermann: In do_query, one could use Service.search_sync instead of Collection.search_sync.
But in do_store, Item.create_sync relies on this `c` being the right collection (with the given label), which is not necessarily true for a user with more than one collection. So please keep this check in there. (Actually there can be more than one liferea collections if anyone else would be using this label for their collection, which I do not assume.)

If you would not care about any old liferea keyrings, you would make use of aliases, which give you a simplified API, but I thought backwards compatibility is better.

Leiaz: I was talking about the `do_deactivate`.

bgermann: I did not touch it, because I don't know any details about the peas plugin system. Maybe it is triggered. If you like, you can remove it after merging.

lwindolf:
Merged and I did the renaming on master. All should be fine. Thanks again!

PR #679, opened 2018-10-15
Let normal view have an horizontal scrollbar, Fixes #549

### LINKED ISSUE #549: Scrollbars not always present in the headline area in Liferea 1.12+
When I tried out the latest git tip, with the date of items displayed on the right, one of the first things I tried was shrink the amount of space given to the date (by grabbing its divider bar and moving it right). When I moved it far enough right, the scrollbar disappeared entirely, even if I hovered the mouse over the very right side of the window. Widening the date area again sometimes but not always seemed to fix this, and at least once I switched over to my 'Important Items' virtual folder and had the scroll bar not there again, even after I'd made it reappear in the feed where I'd first tried shrinking the date field.
  
(Hopefully this description of what I'm seeing is comprehensible and clear enough.)

Leiaz:
This fixes #549 by having an horizontal scrollbar in the item list in normal view, that way the widget size doesn't grow "out" of the window and the vertical scrollbar stays on the side.

@lwindolf I don't know if you want to solve it this way, but I'm posting it here in case people want to use it while waiting for a better solution.

### DIFF
--- a/src/ui/item_list_view.c
+++ b/src/ui/item_list_view.c
@@ -801,10 +801,7 @@ item_list_view_create (gboolean wide)
 	g_object_ref_sink (ilv->priv->ilscrolledwindow);
 	gtk_widget_show (ilv->priv->ilscrolledwindow);
 
-	if (wide)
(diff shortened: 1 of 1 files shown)


lwindolf:
I agree having the scrollbar is the best way to be safe even though it eats some vertical space.

PR #680, opened 2018-10-17
Added xdg-email command in list item popup menu

poetsmeniet:
Please review the changes, this is just a quick modification and I will probably want to add some more checks for robustness. I tried to keep it in style, but please correct me where I went wrong

Also a question: in popup_menu.c there is this line:
 g_menu_item_set_label (menu_item, _("Open In _Tab"));

What does the underscore in "_Tab" denote?

### DIFF
--- a/src/item.c
+++ b/src/item.c
@@ -190,15 +190,13 @@ item_make_link (itemPtr item)
 	return link;
 }
 
-gchar *
+const gchar *
 item_get_author(itemPtr item)
 {
-	const gchar	*src;
-	gchar		*link, *author;
+	gchar *author;
 
 	author = metadata_list_get (item->metadata, "author");
-
-    return author;
+	return author;
 }
 
 void
--- a/src/item.c
+++ b/src/item.c
@@ -190,6 +190,17 @@ item_make_link (itemPtr item)
 	return link;
 }
 
+gchar *
+item_get_author(itemPtr item)
+{
+	const gchar	*src;
+	gchar		*link, *author;
+
+	author = metadata_list_get (item->metadata, "author");
+
+    return author;
+}
+
 void
 item_unload (itemPtr item) 
 {
--- a/src/ui/liferea_shell.c
+++ b/src/ui/liferea_shell.c
@@ -1020,16 +1020,16 @@ email_the_author(GSimpleAction *action, GVariant *parameter, gpointer user_data)
 
 		g_assert (author != NULL);
 
-		argv[0] = g_strdup("xdg-email");
-		argv[1] = g_strdup_printf ("mailto:%s", author);
-		argv[2] = g_strdup("--subject");
-		argv[3] = g_strdup_printf ("%s", subject);
+		argv[0] = g_strdup("xdg-email\0");
+		argv[1] = g_strdup_printf ("mailto:%s\0", author);
+		argv[2] = g_strdup("--subject\0");
+		argv[3] = g_strdup_printf ("%s\0", subject);
 
 		g_spawn_async (NULL, argv, NULL, G_SPAWN_SEARCH_PATH, NULL, NULL, NULL, &error);
 
 		if (error && (0 != error->code)) {
-			debug2 (DEBUG_GUI, "Mail client command failed: %s : %s", argv[0], error->message);
-			liferea_shell_set_important_status_bar (_("Browser command failed: %s"), error->message);
+			debug2 (DEBUG_GUI, "Email command failed: %s : %s", argv[0], error->message);
+			liferea_shell_set_important_status_bar (_("Email command failed: %s"), error->message);
 			g_error_free (error);
 		} else {
 			liferea_shell_set_status_bar (_("Starting: \"%s\""), argv[0]);
--- a/src/ui/liferea_shell.c
+++ b/src/ui/liferea_shell.c
@@ -1000,13 +1000,56 @@ on_action_copy_link_to_clipboard (GSimpleAction *action, GVariant *parameter, gp
 
 }
 
+static void
+email_the_author(GSimpleAction *action, GVariant *parameter, gpointer user_data)
+{
+    itemPtr item = NULL;
+    item = itemlist_get_selected ();
+
+	if (item) {
+        gchar *author, *subject;
+	    GError		*error = NULL;
+	    gchar 		*cmd, **argv;
+	    gint 		argc;
+
+        author = item_get_author(item);
+        subject = item_get_title (item);
+
+        g_assert (author != NULL);
+
+        cmd = malloc(255 * sizeof(char));
+        sprintf(cmd, "xdg-email mailto:%s --subject '%s'", author, subject);
+
+        g_shell_parse_argv (cmd, &argc, &argv, &error);
+
+        if (error && (0 != error->code)) {
+            liferea_shell_set_important_status_bar (_("Browser command failed: %s"), error->message);
+            debug2 (DEBUG_GUI, "Mail client command is invalid: %s : %s", cmd, error->message);
+            g_error_free (error);
+        }
+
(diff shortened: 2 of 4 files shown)


lwindolf:
@poetsmeniet The _ denotes the hotkey when in the menu. In earlier GTK releases it was displayed as an underlined letter. Nowadays this GTK behaviour is turned off in most desktop distributions. So its mostly ineffective.

src/item.c
```diff
@@ -190,6 +190,17 @@ item_make_link (itemPtr item)
 	return link;
 }
 
+gchar *
+item_get_author(itemPtr item)
+{
+	const gchar	*src;
```
Leiaz: You are not using the `src` and `link` variables.

src/item.c
```diff
@@ -190,6 +190,17 @@ item_make_link (itemPtr item)
...
+	gchar		*link, *author;
+
+	author = metadata_list_get (item->metadata, "author");
+
+    return author;
```
Leiaz: There are indentation problems. Liferea uses tabs.

src/item.c
```diff
@@ -190,6 +190,17 @@ item_make_link (itemPtr item)
 	return link;
 }
 
+gchar *
```
Leiaz: I think this should return a `const gchar *`. `metadata_list_get` returns const.

src/ui/popup_menu.c
```diff
@@ -116,6 +117,12 @@ ui_popup_item_menu (itemPtr item, guint button, guint32 activate_time)
 	g_menu_item_set_action_and_target (menu_item, "app.open-link-in-external-browser", "s", item_link);
 	g_menu_append_item (section, menu_item);
 
+    if(author){
+        g_menu_item_set_label (menu_item, _("Email The Author"));
+	    g_menu_item_set_action_and_target (menu_item, "app.email-the-author", "s", item_link);
```
Leiaz: You are passing the item link as parameter to the action's callback, .maybe you intended to pass the author?  But then you are not actually using the parameter in your callback, as it acts on the selected item, and you also use the item's title. Either your action should take both strings as parameters, or the item's id, or nothing (preferably, I think it should use the item passed to this function).

src/ui/liferea_shell.c
```diff
@@ -1000,13 +1000,56 @@ on_action_copy_link_to_clipboard (GSimpleAction *action, GVariant *parameter, gp
...
+	    GError		*error = NULL;
+	    gchar 		*cmd, **argv;
+	    gint 		argc;
+
+        author = item_get_author(item);
+        subject = item_get_title (item);
+
+        g_assert (author != NULL);
+
+        cmd = malloc(255 * sizeof(char));
```
Leiaz: Don't use malloc, use Glib's string functions instead, like `g_strdup_printf`.

Leiaz: In fact, you shouldn't create a command to then parse it in an array : directly create an argument array. Otherwise you have to properly escape special characters in the title.

(Also, if you are new to C, you have a buffer overflow here. Sprintf doesn't take the size of the buffers into account, and you allocate a fixed number of characters without checking which size your final string will be.)

src/ui/popup_menu.c
poetsmeniet: Clear.. I see now that these different ways of passing arguments are all used so will go with you suggestion and only pass the item->id.

Do you require a reply to all comments in this pull request?

Leiaz: No, actual replies are not necessary. Fixing the issues in a commit is enought.

src/ui/liferea_shell.c
poetsmeniet: I pushed my latest changes. Regarding the buffer overflow, I was aware of that one.. :) It is now replaced by Glib function.

The only thing I did not quite get is your remark regarding directly creating an argument array. Special characters are now parsed as literal by using single quotes around the subject. I used the code from execute external browser as an example to implement the xdg-email command.

src/ui/liferea_shell.c
```diff
@@ -1003,42 +1003,48 @@ on_action_copy_link_to_clipboard (GSimpleAction *action, GVariant *parameter, gp
...
+		if (error && (0 != error->code)) {
+			debug2 (DEBUG_GUI, "Mail client command failed: %s : %s", cmd, error->message);
+			liferea_shell_set_important_status_bar (_("Browser command failed: %s"), error->message);
+			g_error_free (error);
+		} else {
+			liferea_shell_set_status_bar (_("Starting: \"%s\""), cmd);
+		}
 
-        g_free(cmd);
+		free(cmd);
```
Leiaz: The `g_free` was correct.

src/ui/liferea_shell.c
```diff
@@ -1003,42 +1003,48 @@ on_action_copy_link_to_clipboard (GSimpleAction *action, GVariant *parameter, gp
...
 
-	if (item) {
-        gchar *author, *subject;
-	    GError		*error = NULL;
-	    gchar 		*cmd, **argv;
-	    gint 		argc;
+	if (parameter)
+		item = item_load (g_variant_get_uint64 (parameter));
+	else
+		item = itemlist_get_selected ();
```
Leiaz: Those two functions allocate a new item. You must use `item_unload` to free it.

src/ui/liferea_shell.c
```diff
@@ -1003,42 +1003,48 @@ on_action_copy_link_to_clipboard (GSimpleAction *action, GVariant *parameter, gp
...
-        } else {
-            liferea_shell_set_status_bar (_("Starting: \"%s\""), cmd);
-        }
+		if (error && (0 != error->code)) {
+			liferea_shell_set_important_status_bar (_("Browser command failed: %s"), error->message);
+			debug2 (DEBUG_GUI, "Mail client command is invalid: %s : %s", cmd, error->message);
+			g_error_free (error);
+		}
+
+		g_spawn_async (NULL, argv, NULL, G_SPAWN_SEARCH_PATH, NULL, NULL, NULL, &error);
```
Leiaz: You are calling spawn even if there was an error when parsing.

src/ui/liferea_shell.c
poetsmeniet: I see, made some changes. Please advise if there is a more graceful solution to filling up argv

src/ui/liferea_shell.c
poetsmeniet: I made modifications, I hope the manual filling of argv is ok?

Leiaz: `argv` must be NULL terminated. Except for that it's good.

poetsmeniet: OK cool

poetsmeniet:
> Sorry, that's not what I meant. String literals are already null terminated. I meant the last element of `argv` should be `NULL`, to mark the end of the array. That is how the array can be.used without having to pass the size along with it to functions.

No really, my bad.. should have read your comment more thoroughly. Should be fixed now..

lwindolf:
@Leiaz would you say this is ready for merge?

Leiaz:
I would say yes, but I wanted to compile and test it before confirming and haven't been able to. I have been away from my desktop for a while, I'm staying with family, helping with family business paperwork that my grandma used to do on a typewriter and countless little books and pieces of paper on an antique Windows laptop (which must run Windows and antique Microsoft Office for quasi religious reasons, aargh :). I could be away from a usable computer for some more time.
So if you tested and didn't see something broken I could have missed, I'd say it is ready 👍

lwindolf:
@Leiaz Your task sounds otherworldy, when described on github :-) Still I believe it's worth doing for family! Have fun and don't worry, I'll test the PR

lwindolf:
@poetsmeniet Looks good. Merged.

PR #686, opened 2018-10-27
Confirmation on adding duplicate url source (resolves #658)

### LINKED ISSUE #658: Avoid duplication of subscription
From my experience, liferea will add another subscription even if I already subscribed to a given address. Would be great to allow this.

dymoksc:
#658

### DIFF
--- a/src/ui/feed_list_node.c
+++ b/src/ui/feed_list_node.c
@@ -427,3 +427,48 @@ feed_list_node_remove (nodePtr node)
 	g_signal_connect (G_OBJECT (dialog), "response",
 	                  G_CALLBACK (feed_list_node_remove_cb), node);
 }
+
+static void
+feed_list_node_add_duplicate_url_cb(GtkDialog *dialog, gint response_id, gpointer user_data)
+{
+	if (GTK_RESPONSE_ACCEPT == response_id)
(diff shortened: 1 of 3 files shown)


src/ui/feed_list_node.c
```diff
@@ -427,3 +427,56 @@ feed_list_node_remove (nodePtr node)
...
+
+void
+feed_list_node_add_duplicate_url_subscription (subscriptionPtr tempSubscription, nodePtr exNode)
+{
+	GtkWidget	*dialog;
+	GtkWindow	*mainwindow;
+	gchar		*text;
+
+	text = g_strdup_printf (
+			_("Are you sure that you want to add a tempSubscription with URL \"%s\"? Another subscription with the same URL already exists (\"%s\")."),
```
lwindolf: I think you accidentily left "tempSubscription" in the text. Probably "a new subscription" is what is intended.

dymoksc: Actually, it was intentional. I was trying to emphasize, that it's not yet the subscription struct, which is going to be added to user subscriptions, since it's not passed as whole and is later reconstructed in `feedlist_add_subscription()`. But if it's not that obvious, how I imagined, I have no problems with renaming it:)

dymoksc: Oh, you was talking about the string...

Issue #685, opened 2018-10-27
The "Mark All As Read" button in the headerbar plugin is not feed-specific.


rolandixor:
I know I was the one who requested the button, but it doesn't do what the non-headerbar version does - it marks all feeds as read, as opposed to the one currently selected.

PR #687, opened 2018-10-30
Change headerbar mark-read button to act on selected feed, not all feeds

cooperra:
Addresses #685, which is about consistency with the non-headerbar version of the button.


lwindolf:
Thanks for providing the fix!

PR #688, opened 2018-10-31
Add tooltips to headerbar buttons, with translations

cooperra:
We're piggybacking off of Liferea's existing translations to make this work.

If they change, it's possible this will break and fallback to English.

Also, some of these have mnemonics defined (with "_") which we remove for our purposes.


lwindolf:
Great improvement! I didn't know so far on how to create translatable plugins.

lwindolf:
From a distribution perspective I'd say it's perfect for included plugins to be translated via the central .po files

PR #689, opened 2018-11-09
Fix feed update failures after suspend

poetsmeniet:
Tested this on Debian 9 with NM 1.6.2-3. The modification will allow for updating feeds only with Site or Global connectivity


lwindolf:
Thanks a lot for this important fix for laptop users!

PR #699, opened 2018-12-10
Duplicates check moved form subscription_dialog.c to the separate function

dymoksc:
Continues #686 
@lwindolf, you are of course right about the unnecessary code duplication, but `feedlist_add_subscription()` is used in multiple different context, about which I'm not 100% sure. So, I've moved that logic into a separate function `feedlist_add_subscription_check_duplicate()`, which, in its turn, calls `feedlist_add_subscription()`. However, if you are sure, that opening confirmation dialog will be appropriate in all the context, from which `feedlist_add_subscription()` is called, I'll gladly move it there


lwindolf:
Thank you for refactoring this!

PR #705, opened 2019-01-26
Action to open enclosures

### LINKED ISSUE #700: Add hotkey to open attachments
So, I'd like to be able to open attachments using the keyboard. For example, I have set "mpv" and "feh" to open video and image attachments, but having to use the mouse to do it is cumbersome..

Would something like this be easy to implement?

Leiaz:
This adds an action to open the next enclosure in the list, with the shortcut Ctrl+o. Fixes #700.

### DIFF
--- a/src/ui/enclosure_list_view.c
+++ b/src/ui/enclosure_list_view.c
@@ -337,6 +337,37 @@ enclosure_list_view_select (EnclosureListView *elv, guint position)
 }
 
 void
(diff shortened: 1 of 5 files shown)


lwindolf:
I'm not sure if I'm doing something wrong, but Ctrl+o isn't triggering... :-(

lwindolf:
Tested once more and it did work!

PR #710, opened 2019-03-18
Webkit Widget : replace deprecated GdkScreen functions for font size

Leiaz:
This replaces deprecated gdk_screen_get_width/height.

I connect both the "realize" and "screen-changed" signals because "screen-changed" is emitted when the widget doesn't have a GdkWindow set yet (is not realized yet). The idea is to handle both setting the font at boot and when changing screen when running.

I have only one low dpi screen so I didn't test completely.

### DIFF
--- a/src/webkit/webkit.c
+++ b/src/webkit/webkit.c
@@ -1,7 +1,7 @@
 /**
  * @file webkit.c  WebKit2 browser module for Liferea
  *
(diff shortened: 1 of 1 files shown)


lwindolf:
Thanks for the refactoring and cleanup!!! 

The Travis-CI Dockerfile is updated now so builds work with the new dependencies.

PR #714, opened 2019-03-22
Replaces deprecated gtk_menu_popup

Leiaz:
This replaces the deprecated `gtk_menu_popup` function with `gtk_menu_popup_at_pointer`. The behavior appears to be the same as in master. But maybe we should handle the menus from the "popup-menu" signal (Shift + F10) differently : currently it appears where the pointer is but is not related to what is under the pointer.

### DIFF
--- a/src/ui/enclosure_list_view.c
+++ b/src/ui/enclosure_list_view.c
@@ -97,7 +97,7 @@ enclosure_list_view_get_selected_enclosure (EnclosureListView *elv, GtkTreeIter 
 }
 
 static gboolean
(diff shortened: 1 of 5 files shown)


lwindolf:
I checked Nautilus (popup at focussed widget), Firefox (popup always top left) and GnomeTerminal (popup at mouse) for reference and for me they all behave randomly with Shift + F10. As a user I cannot see a pattern where it the menu is "normally" supposed to pop up.

So I think going the easy way similar to Nautilus might not be that bad.

Thanks again for your great refactorings!!!

PR #734, opened 2019-07-31
Fixes crash when selecting empty nodes

Leiaz:
The dummy empty "nodes" created in `src/ui/feed_list_view.c` in order to
make empty folders expandable are causing problems because they have a
NULL node pointer. The "selection-changed" callback was expecting a non
NULL node pointer.
When an empty node is selected I get the parent's node and advertise it
as selected. The empty node being a placeholder for the folder, this
results in the proper behavior when the empty node is selected and a
feed is added.

I'm not sure this is the best way to fix this. Those empty nodes are a
special case in our GtkTreeStore, and there may be other places where
they need to be handled differently from other nodes.


lwindolf:
The solution makes a lot of sense to me!

Sorry for the lag in reviews recently. Lot of things going on in family life which take precedence :-)

lwindolf:
Commit message wrong. Not master, but liferea-1_12 branch was the merge target, so it is included in 1.12.7

PR #742, opened 2019-09-18
Consider startup subscriptions update an auto-update

### LINKED ISSUE #730: "Update all subscriptions at startup" overrides "Don't update this feed automatically"
I have some feeds where I'd like to keep a copy of all the posts (so I don't want to delete the feed), but the feed domain has been taken over by spammers and when the feed gets updated in liferea, I receive spam posts.

I can disable automatic updates to the spam feeds but I also have "Update all subscriptions at startup" enabled and when I start liferea, then the spam feeds get updated too.

The startup feed refresh is an automatic refresh, so it should also get disabled for the feeds where automatic updates are disabled.

dmytrokyrychuk:
Resolve #730


lwindolf:
Thanks a lot for debugging and fixing this!

PR #745, opened 2019-09-23
Update formats-ouverts.org feed

sblondon:
`http://formats-ouverts.org/blog/` returns a 404. It seems the blog is now displayed at `/`.
`http://formats-ouverts.org/rss.php` redirects to `http://formats-ouverts.org//feed/atom`.

This PR updates those links.


lwindolf:
Thanks for the fix!

PR #747, opened 2019-09-25
Update liferea.css with better fonts

AzharMithani:


lwindolf:
Thank you for taking the time to create this change. While it looks reasonable can you give some indication on the why? It's hard to test without a rationale :-)

PR #749, opened 2019-09-29
Readability.js to format AMP/HTML5 fetched content

lwindolf:
This PR adds Readability.js which really improves formatting of stripped AMP/HTML5 content.
- Fixes "loose bullets" and overly large whitespace ranges.
- For now the idea is to have it always on, but it could be disabled when there is no AMP/HTML5 content
- Dropping 2-pane mode as it is incompatible with Readability.js concept of convert 1 DOM to 1 article


css/liferea.css
```diff
@@ -73,7 +73,7 @@ dd {
 	padding-left: 30px;
 }
 
-section, video, svg, iframe, figure, div.content * img {
```
lwindolf: iframe was removed here for as Readability drops it anyway

Leiaz:
I'm testing this. I haven't had the time to look in detail at the code (and for the javascript I'm not really the right person to ask).

About dropping combined view : I almost never used it because when you have a lot of feed items, many linking to at least one image, there is a huge spike in bandwidth usage as all the images are fetched. That looked like the sort of automated download that is not very nice to the website owner. So I don't mind it being removed.

I usually use Liferea with javascript disabled, as it should not be necessary for what I want to see. With javascript enabled, some items have autoplaying videos, I guess some people consider this to be a feature ... :D Currently disabling javascript also disable any javascript we may add. But we'll have to find a way to fix this anyway as some [stuff I was using in the WebKit web extension for the article scrolling is being deprecated](https://webkitgtk.org/reference/webkitdomgtk/stable/WebKitDOMDOMWindow.html).

Leiaz:
In WebKit's console I'm getting : 
```
[Error] TypeError: null is not an object (evaluating 'documentClone.getElementById('shading').innerHTML')
	documentIsReady (liferea://:6:959)
```
There is a div with class "shading" but no id. That div should probably be given an id.

lwindolf:
Good hint on the JS dependencies. I haven't thought on the relationship to the disable JS feature. 

There needs to be at least a warning about both features being incompatible. Or alternatively a guarantee that a Readability.js displayed item does not use external Javascript. E.g. like renaming the feature to "Never run external Javascript"

nekohayo:
OMG YES thank you so much for working on this! I wanted to request it but had not dared to, for I know there is a "no feature requests" policy that I try to respect :)

If it can be of any help at all, Epiphany implemented this (just like GNOME News) some years ago, and you can find all the details/code/gotchas there: https://bugzilla.gnome.org/show_bug.cgi?id=772831

To me, the reader mode is an opportunity to read the contents without being limited by the abstract, but without loading all the navigation and javascript crap of news websites. If there can be a way for Liferea to offer it as a "default method of browsing" (i.e. be able to have the reader mode activated by default globally and/or per-feed) this would be amazing.

lwindolf:
> In WebKit's console I'm getting :
> 
> ```
> [Error] TypeError: null is not an object (evaluating 'documentClone.getElementById('shading').innerHTML')
> 	documentIsReady (liferea://:6:959)
> ```
> 
> There is a div with class "shading" but no id. That div should probably be given an id.

@Leiaz Should be fixed now

PR #752, opened 2019-10-11
Fix #693 - Add option to disable minimizing to tray.

### LINKED ISSUE #693: How to not minimize to tray?
I don't like the default behaviour of liferea, which is to minimize to tray when closing. Is there a way to close it instead of minimizing it?

BurnhamG:
This closes #693 by extending the Trayicon plugin, allowing the user to choose what Liferea does when closing the main window. Let me know if you have any questions, or would like anything changed.

### DIFF
--- a/doc/html/faq_en.html
+++ b/doc/html/faq_en.html
@@ -68,12 +68,11 @@ do work.</dd>
 </dd>
 
 <dt>Q: Liferea is buggy! It does not close when I click on the window
-managers close button and the tray icon is activated.</dt>
+manager's close button and the tray icon is activated.</dt>
 <dd>A: 
 This is a pretty disputed topic. We see this as a useful feature, other
-people don't. If you don't like the behaviour you can change it in the
-preferences under "GUI". There you can enable the check box "Terminate
-instead of minimizing to tray icon".
+people don't. If you don't like the behaviour you can change it by right-clicking 
+the tray icon and toggling the checkbox labelled "Minimize to tray on close."
 </dd>
 
 <dt>Q: I want cookies!</dt>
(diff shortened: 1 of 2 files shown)


lwindolf:
Awesome contribution that a lot of users will like. Will have a look at it soon!

SukkoPera:
This is nice but has a small bug where the actual value of the setting is not read at startup, so even if minimize to tray is disabled, it shows up as enabled in the tray icon context menu. In this case, if the app is closed, it will exit.

BurnhamG:
Thanks for reporting that, @SukkoPera. I'll take a look ASAP.

BurnhamG:
The bug should be resolved now - apologies for that oversight, and thank you for pointing it out.

Leiaz:
The first time it is used, with no config file or folder : 
```
  File "/home/leiaz/local-inst/lib/liferea/plugins/trayicon.py", line 107, in do_activate
    self.config_path = self.get_config_path()
  File "/home/leiaz/local-inst/lib/liferea/plugins/trayicon.py", line 155, in get_config_path
    self.save_config("True")
  File "/home/leiaz/local-inst/lib/liferea/plugins/trayicon.py", line 174, in save_config
    with open(self.config_path, "w") as f:
AttributeError: 'TrayiconPlugin' object has no attribute 'config_path'
```
And the menu is empty.
On second run it works.

Also, when quitting : 
```
  File "/home/leiaz/local-inst/lib/liferea/plugins/trayicon.py", line 244, in do_deactivate
    self.window.disconnect_by_func(self.trayicon_close_behavior)
TypeError: nothing connected to <bound method TrayiconPlugin.trayicon_close_behavior of <trayicon.TrayiconPlugin object at 0x7fff928550c0 (TrayiconPlugin at 0x555555824e20)>>
```

lwindolf:
I'm merging this, so we won't loose this often requested contribution. 
Will have a look into the CLI visible warnings next.

lwindolf:
@Leiaz @BurnhamG I believe 5e1f9f7 fixes the issues @Leiaz reported. Please review!

Leiaz:
@lwindolf Yes, it works.

BurnhamG:
@lwindolf Just for a second data point, 5e1f9f7 fixes the reported issues for me as well. Thanks for the patch and merge!

lwindolf:
Thanks for the feedback!

PR #764, opened 2019-12-16
Extend MediaRSS support

mikelolasagasti:
Basic support for MediaRSS to include video description and thumbnail.

#761 

Signed-off-by: Mikel Olasagasti Uranga <mikel@olasagasti.info>

PR #765, opened 2019-12-25
Embed YouTube videos

mikelolasagasti:
Extend xslt to render YouTube's embeded code if source is a YouTube video.

Another option would be to check the YT namespace, but adding another parser
might be overkill. Also, I was not able to find the YT schema.

Privacy wise, there would be a request everytime a feed is rendered, but this
should only happen if the feed is a YT MediaRSS feed or if the feed has a YT
link as main link. If description has a YT url, the embed code won't show.

Signed-off-by: Mikel Olasagasti Uranga <mikel@olasagasti.info>

### DIFF
--- a/xslt/item.xml.in
+++ b/xslt/item.xml.in
@@ -4,6 +4,7 @@
 /**
  * Rendering stylesheet for Liferea (item view: item rendering)
  *
+ * Copyright (C) 2009-2019 Mikel Olasagasti Uranga <mikel@olasagasti.info>
  * Copyright (C) 2006-2019 Lars Windolf <lars.windolf@gmx.de>
  *
  * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


mikelolasagasti:
@lwindolf the embed iframe works fine, but when the video is clicked, it is moved to a different position and buttons don't work (unless one guess where to click). 

I checked the generated xhtml file on a browser and works fine. 

I understand this is an issue with WebkitGTK, right? If this is the case, then this code won't be very useful.

mikelolasagasti:
@lwindolf I tested with current master and now videos work fine.

You can test with this source for example: https://www.youtube.com/feeds/videos.xml?channel_id=UC7_gcs09iThXybpVgjHZ_7g

This would extend an complete already merged PR https://github.com/lwindolf/liferea/pull/764

lwindolf:
Works perfectly!

PR #773, opened 2020-01-25
enable WebKit WebInspector

sillyslux:
adding "Inspect" in web view context menu

### DIFF
--- a/src/webkit/liferea_web_view.c
+++ b/src/webkit/liferea_web_view.c
@@ -20,6 +20,7 @@
 
 #include "liferea_web_view.h"
 
+#include "../debug.h"
 #include "browser.h"
 #include "common.h"
 #include "enclosure.h"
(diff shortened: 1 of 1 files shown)


sillyslux:
This introduces a new issue:
Global hotkeys are interfering e.g. when you type in the js-console.
To work around you can click the "detach" button to move the inspector to a separate window, once it's shown.
There's also [a detach method](https://webkitgtk.org/reference/webkit2gtk/2.7.1/WebKitWebInspector.html#webkit-web-inspector-detach), but it should be called only after the inspector is ready. I suspect this has to be done async, from a separate thread.

lwindolf:
Nice helpful debugging feature. Also good hint on the hotkey intercepting workaround. Didn't know this, but makes perfectly sense!

sillyslux:
I really think the interfering hotkeys should be disabled during debugging or alternatively the inspector automatically detached with the aforementioned detach method.

Also mentioning this in docs and the manpage might be nice. like
`.B \-\-debug\-html
Enables HTML rendering debugging. Each time Liferea
renders HTML output it will also dump the generated
HTML into $XDG_CACHE_DIR/liferea/output.xhtml.
Enables WebInspector for WebView (in context menu).`

...and, sadly, it should also be noted that the debugger window should be detached.

Do you want me to?

lwindolf:
@sillyslux Please feel free to do so. I think this is very helpful for contributors

PR #776, opened 2020-01-26
Theme reload in WebView without restart

sillyslux:
the first commit allows users to send SIGHUP to invoke the css rebuild, later i've remembered [the style-updated signal](https://developer.gnome.org/gtk3/stable/GtkWidget.html#GtkWidget-style-updated).
Now there's both, i feel like that's redundant...
otoh, SIGHUP might still be useful when someone modifies ~/.config/hexchat/liferea.css


lwindolf:
@sillyslux Sorry for taking so long to test this properly!

Actually I cannot get the CSS rule reload get to work without restarting when changing the theme in Ubuntu 20.04 with Gnome Tweaks. Do you have any idea what I might be doing wrong?

sillyslux:
i'm using `gsettings set org.gnome.desktop.interface gtk-theme Adwaita` and `gsettings set org.gnome.desktop.interface gtk-theme Adwaita-dark` in a script.
i'll try it with gnome and gnome tweaks and report back...

lwindolf:
@sillyslux Thanks for the hint. I tried with `gsettings set org.gnome.desktop.interface gtk-theme Adwaita-dark` too and it also did not trigger an CSS change for me.

sillyslux:
just to be clear... the gtk interface switches from light to dark, but the webkit frame does not?
did you run it with --debug? did you see the output from color detection?
in webkits inspector, the css link src, does it have a cachebuster path? (like href="file:///home/silly/.cache/liferea/style.css?1595446640073" <- the ?timestamp)

Leiaz:
With the same theme, I get a darker background for the number of unread items with this PR than without, for some reason.
Liferea_without_theme_patch_2020-07-23_19-12-38
Liferea_with_theme_patch_2020-07-23_19-10-34

Otherwise theme switching works for me. I'm using Xfce and the "Appearance" tool. Theme switching takes ages for some reason (globally, not just Liferea). And most of the themes I have aren't really usable but that is another issue :)
CSS is reloaded directly with javascript enabled, otherwise the change is visible on next item selection.

lwindolf:
@sillyslux Sorry for taking so long to reply!

Yes, I 
- see a correct cache-buster path in the CSS href
- also have all GTK widgets changing the theme color without issues
- am also changing to another item to avoid any JS issues
- do not see any debug output on the color stuff

I added a debug print in liferea_shell_rebuild_css() and it is never triggered. It is very very strange.

src/ui/liferea_shell.c
```diff
@@ -1259,6 +1259,7 @@ liferea_shell_create (GtkApplication *app, const gchar *overrideWindowState, gin
 	gtk_widget_show_all (shell->statusbar_feedsinfo_evbox);
 	gtk_box_pack_start (GTK_BOX (shell->statusbar), shell->statusbar_feedsinfo_evbox, FALSE, FALSE, 5);
 	g_signal_connect (G_OBJECT (shell->statusbar_feedsinfo_evbox), "button_release_event", G_CALLBACK (on_next_unread_item_activate), NULL);
+	g_signal_connect (G_OBJECT (shell->statusbar), "style-updated", G_CALLBACK(liferea_shell_rebuild_css), NULL);
```
lwindolf, line 1262: Found the issue. I'm a user of the "floating statusbar plugin", which hides the status bar when it is unused to save vertical space. And registering the signal to a widget that is not always visible seems to be the problem.

If I register to the shell->window the signal works properly for me. 

I'm guessing you are not using the plugin :-)

Still let's register the signal to "shell->window" to eliminate this corner case.

Actually this raised the next corner case: the main window being invisible (when hidden to tray icon). Then the signal is also lost...

lwindolf (changes requested):

src/ui/liferea_shell.c
sillyslux: I remember thinking of this and that i went looking for a "View -> Hide Statusbar" option in the menu.

sillyslux: I think i chose something other than the main window for a reason, also i think i've found the issue with the slighter darker badges...

lwindolf: I hope you don't mind me moving the signal for now to the mainwindow. If you come across the reason for not having it there please raise a bug ticket so we can follow up on it. For now I want to have this great PR merged for 1.13.3 :-)

PR #783, opened 2020-02-24
Update IS_STATE when update item in itemlistview

mozbugbox:
The state should be kept consistance. The state can be changed by user.


lwindolf:
You are totally right!

PR #785, opened 2020-02-24
Update French Translation

guilieb:
Hi,

Here’s an update for the French Translation on the 1.12 branch 😉


Leiaz:
I merged this on master.
@lwindolf Should it also be merged on 1.12 ? I tried it on 1.12, but there is likely another bug that was fixed since on master, because I don't see the translations in the menu bar and toolbar.

lwindolf:
@Leiaz Might be good to have on 1.12 too.

PR #800, opened 2020-03-10
Expose remove-item action

mozbugbox:
Remove an item by id without select it.

### DIFF
--- a/src/ui/liferea_shell.c
+++ b/src/ui/liferea_shell.c
@@ -904,7 +904,6 @@ static const GActionEntry liferea_shell_gaction_entries[] = {
 	{"export-feed-list", on_menu_export, NULL, NULL, NULL},
 	{"quit", on_menu_quit, NULL, NULL, NULL},
 	{"remove-selected-feed-items", on_remove_items_activate, NULL, NULL, NULL},
-	{"remove-item", on_action_remove_item, "t", NULL, NULL},
 	{"prev-read-item", on_prev_read_item_activate, NULL, NULL, NULL},
 	{"next-read-item", on_next_read_item_activate, NULL, NULL, NULL},
 	{"next-unread-item", on_next_unread_item_activate, NULL, NULL, NULL},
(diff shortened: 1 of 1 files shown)


Leiaz:
> Somehow action in liferea_shell_add_gaction_entries[] don't activate

It was in `liferea_shell_item_gaction_entries` and the group exists specifically to disable all actions inside it together when no item is selected. The actions are for use in buttons and menus, we want them to be greyed out, that is disabled, when they are not usable. 
There are five action groups `generalActions`, `addActions`, `feedActions`, `itemActions`, and `readWriteActions`. The general actions are always enabled. The feed actions are enabled when there is a selected feed, the item actions are enabled when there is a selected item. The add actions are enabled when it is possible to create new subscriptions, folders, etc ... that is, when the selected node's source node has `NODE_SOURCE_CAPABILITY_WRITABLE_FEEDLIST`. The read write actions are the same, except they are disabled when an "empty" node inside an empty folder is selected : adding is allowed but as it is not a real node, it can't be removed.

Leiaz:
> The variantType of copy-item-to-newsbin somehow crash glib

All actions that are added to the GApplication are automatically made available over DBus. The signature (umt) is not a valid DBus signature because it contains a "maybe". There are other actions that should probably not be added to the application either as it doesn't really make sense to have them available from outside.

mozbugbox:
Well, the actions can be used in plugins which can supply parameters to actions.

Like when I want to remove duplicated items in a itemlistview, I can use remove-item directly. Otherwise, I have to move the cursor to the item. Then call remove-selected-item, and I have to restore the unread status of the next item auto-selected once the previous item got removed.

Leiaz:
I was talking about the fact that they are available over DBus.
When I wrote it, the suggestion in the doc seemed to be to have a "win" group for windows actions and "app" group for other application actions. It makes sense that "open" actions and the like should be available from outside, perhaps "about" and "application preferences" too, but for many of our actions it doesn't really make sense to call them from outside Liferea. They could be in another group, not added to "app" and not made automatically available over DBus.

lwindolf:
@mozbugbox @Leiaz I'm unsure how to proceed on this PR :-)

Leiaz:
We could perhaps create a specific group of actions to make available to plugins. And have the group as a property of LifereaShell or available with a getter function ... It wouldn't be added to the app so we could have parameters not necessarily compatible with DBus ...
It still feels a bit hacky. I'm not sure this is the best way to make some functions accessible to plugins. The "action" part is not really needed, I think ?
Also, how stable are we supposed to keep the stuff "exposed" to plugins in this way ? If we refactor some stuff, plugins could be broken as result. Which is already the case, it's not like we have elements clearly labelled "plugin API, do not break" ...

But this pull request works and doesn't break anything, so it can also be merged as-is ...

Issue #787, opened 2020-03-08
When an unread item is already selected and marked as unread again, "Jump to next unread" button should allow skipping it


nekohayo:
When I jump to the next unread, sometimes I don't want to read that particular item so I mark it as unread again (with the middle click button)... except that I then can't use the "Jump to next unread" button again to skip to the next one.

rich-coe:
This might be a dup of #778

lwindolf:
Like in mail clients there are two use cases

1. empty inbox type (marks everything read, uses flags/newsbins for "read-later")
2. messy inbox type (keeps stuff unread for later)

As I'm of type 1 I design Liferea around it. Supporting both uses cases at the same time is to hard. Especially with "Next Unread" already being a complex feature.

Closing this as won't fix.

PR #821, opened 2020-04-25
Skip current item when finding next unread item

tper0700:
This is a possible fix for #787. When looking for the next item, the code will now skip over the current selected index.

The main fix is in item_list_view.c. If the code hasn't moved past startId, don't return that entry.

The change in itemview.c is needed because itemview_find_unread_item will restart the search from the top of the list, returning to startId which may have been marked unread.


lwindolf:
The change is a good fix. Reopening and merging. I hope this is ok.

tper0700:
Awesome, thanks! Can't wait to see this on my production boxes!

PR #822, opened 2020-04-25
first part fix #519  single update to treeview, lookup; adding item_ids

### LINKED ISSUE #519: Improve time to switch folders
Switching folders can take a long time.
I got poor results trying to identify the issue being called from itemset_foreach().

I see two different approaches that could fix this
(1)  Load as many items as possible into the displayed folder as can be done in 250ms.  Display the folder and load the rest as a background task.
(2)  Create a custom GTK Tree Model to load items.  Initially it loads ~100 (?) items and loads new items as the user scrolls through the list.

rich-coe:
first part fix #519  single update to treeview, lookup; and adding item to item_ids
    - per https://developer.gnome.org/glib/stable/glib-Singly-Linked-Lists.html
      change g_slist_append() to g_slist_prepend() as g_slist_append() seeks to the end of the list
      and is inefficient
      + during batch add, don't lookup item in the empty list
      + item add is always followed by item update, pass the just inserted item id into update internal
    - per https://stackoverflow.com/questions/3547743/gtktreeview-insert-update-performance-penalty-because-of-sorting (3)
      + ref. https://eccentric.one/misc/pygtk/pygtkfaq.html#13.43
      + turn off the sorting
      + freeze child notify
      + batch insert into tree
      + thaw child notify
      + turn on sorting

After changes, folder switch update time reduced 80%, (15s to 2.5s, 48s to 9.48, (large) to ~30s).

### DIFF
--- a/src/ui/item_list_view.c
+++ b/src/ui/item_list_view.c
@@ -417,7 +417,7 @@ item_list_view_clear (ItemListView *ilv)
 	GtkTreeStore		*itemstore;
 	GtkTreeSelection	*select;
 
-	itemstore = GTK_TREE_STORE (gtk_tree_view_get_model (ilv->treeview));
+        itemstore = GTK_TREE_STORE (gtk_tree_view_get_model (ilv->treeview));
 
 	/* unselecting all items is important to remove items
 	   whose removal is deferred until unselecting */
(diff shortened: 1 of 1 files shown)


lwindolf:
Great contribution!!!

lwindolf:
Works very well. I can reproduce time savings of ~30-40% on my setup.

nekohayo:
I haven't tested this, but I just wanted to say thank you @rich-coe for taking the time and effort to work on performance improvements like that, very cool :)

PR #826, opened 2020-05-03
Fix save/restore position issues with tray icon

mhoran:
There were a few issues with tray icon save/restore:

* Fix raising Liferea from another workspace using official GDK API.
* Clicking the tray icon didn't save position. Now we just call toggle.
* Closing or minimizing the window wouldn't save position either.

### DIFF
--- a/plugins/trayicon.py
+++ b/plugins/trayicon.py
@@ -133,7 +133,8 @@ class TrayiconPlugin (GObject.Object, Liferea.ShellActivatable):
             self.window.hide()
 
     def trayicon_click(self, widget, data = None):
-        self.shell.toggle_visibility()
+        # Always show the window on click, as some window managers misbehave.
+        self.shell.show_window()
 
     def trayicon_minimize_on_close(self, widget, event):
         self.shell.save_position()
--- a/plugins/trayicon.py
+++ b/plugins/trayicon.py
@@ -144,7 +144,7 @@ class TrayiconPlugin (GObject.Object, Liferea.ShellActivatable):
         self.shell.toggle_visibility()
 
     def trayicon_quit(self, widget, data = None):
-        Liferea.shutdown()
+        Liferea.Application.shutdown()
 
     def trayicon_popup(self, widget, button, time, data = None):
         self.menu.popup(None, None, self.staticon.position_menu, self.staticon, 3, time)
--- a/plugins/trayicon.py
+++ b/plugins/trayicon.py
@@ -129,13 +129,14 @@ class TrayiconPlugin (GObject.Object, Liferea.ShellActivatable):
         "Hide window when minimize"
         if event.changed_mask & event.new_window_state & Gdk.WindowState.ICONIFIED:
             self.window.deiconify()
+            self.shell.save_position()
             self.window.hide()
 
     def trayicon_click(self, widget, data = None):
-        self.window.deiconify()
-        self.window.show()
+        self.shell.toggle_visibility()
 
     def trayicon_minimize_on_close(self, widget, event):
+        self.shell.save_position()
         self.window.hide()
         return True
 
--- a/src/ui/liferea_shell.c
+++ b/src/ui/liferea_shell.c
@@ -1352,15 +1352,28 @@ liferea_shell_present (void)
 	gtk_window_present (shell->window);
 }
 
+static gboolean
+liferea_shell_window_is_on_other_desktop(GdkWindow *gdkwindow)
+{
+#ifdef GDK_WINDOWING_X11
+	return GDK_IS_X11_DISPLAY (gdk_window_get_display (gdkwindow)) &&
+	    (gdk_x11_window_get_desktop (gdkwindow) !=
+	     gdk_x11_screen_get_current_desktop (gdk_window_get_screen (gdkwindow)));
+#else
+	return FALSE;
+#endif
+}
+
 void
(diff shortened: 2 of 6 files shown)


src/ui/liferea_shell.c
```diff
@@ -1355,23 +1356,15 @@ void
...
-	if (gdk_window_get_state (gtk_widget_get_window (mainwindow)) & GDK_WINDOW_STATE_ICONIFIED) {
-		/* The window is either iconified, or on another workspace */
-		/* Raise it in one click */
-		if (gtk_widget_get_visible (mainwindow)) {
-			liferea_shell_save_position ();
-			gtk_widget_hide (mainwindow);
-		}
+	if (gdk_x11_window_get_desktop (gdkwindow) !=
+	    gdk_x11_screen_get_current_desktop (gdk_window_get_screen (gdkwindow))) {
+		gdk_x11_window_move_to_current_desktop (gdkwindow);
```
Leiaz, line 1363: This is X11 specific.
The [doc suggests](https://developer.gnome.org/gdk3/stable/gdk3-X-Window-System-Interaction.html) to check the backend in use.
I don't have Wayland installed, so I didn't check what happens.
This function is only used in the trayicon plugin for now and the plugin is useless on Wayland, but it shouldn't be specific to this plugin. Or then, we should just move it there, but I think hiding the window might be desirable on Wayland too ?

mhoran: Good catch. I think it'd be easiest to keep this in here vs. the plugin, but I agree it should be wrapped in a check. I also don't have Wayland, so I'm not sure what happens there.

If someone feels strongly that this should be moved into the plugin, I can look into that as well. I agree that hiding the window might be desirable on Wayland, in which case keeping protocol specific bits in here would seem to make the most sense.

mhoran: I went ahead and wrapped the X11 specific APIs as described in the docs.

src/ui/liferea_shell.c
```diff
@@ -1352,15 +1352,28 @@ liferea_shell_present (void)
...
 liferea_shell_toggle_visibility (void)
 {
 	GtkWidget *mainwindow = GTK_WIDGET (shell->window);
 	GdkWindow *gdkwindow = gtk_widget_get_window (mainwindow);
 
-	if (gdk_x11_window_get_desktop (gdkwindow) !=
-	    gdk_x11_screen_get_current_desktop (gdk_window_get_screen (gdkwindow))) {
+	if (liferea_shell_window_is_on_other_desktop (gdkwindow)) {
+#ifdef GDK_WINDOWING_X11
 		gdk_x11_window_move_to_current_desktop (gdkwindow);
```
Leiaz, line 1375: Right now, `liferea_shell_window_is_on_other_desktop` imply `GDK_IS_X11_DISPLAY` but to be future proof, I think it would be better to check here too.

mhoran: Sounds good -- I went back and forth on this one, but I think that's a good call.

plugins/trayicon.py
```diff
@@ -129,21 +129,22 @@ def window_state_event_cb(self, widget, event):
         "Hide window when minimize"
         if event.changed_mask & event.new_window_state & Gdk.WindowState.ICONIFIED:
             self.window.deiconify()
+            self.shell.save_position()
             self.window.hide()
 
     def trayicon_click(self, widget, data = None):
-        self.window.deiconify()
-        self.window.show()
+        self.shell.toggle_visibility()
```
lwindolf, line 136: This change will not work. Actually there should be a large comment there about why (my fault for it being missing)...

The rationale is that forcing showing Liferea on any tray click is the only way to support all the hidden window bug reports we've seen over the last 2 years. There were users who could not access the application anymore because there desktop environment caused the visibility check to fail and the app never showed up again.

While it would be great to have toggling for all the people where it works properly, it cannot be for all those broken setups...

lwindolf:
@mhoran thanks for fixing my liferea_shutdown() mess up!!!

plugins/trayicon.py
mhoran: Got it. I remember this used to be the behavior and had a vague recollection that there was some reason it wasn't this way anymore, but couldn't remember why. I'll leave a comment.

I didn't want to duplicate the logic in here but part of the reasoning for this change is to make sure that we restore the saved window position (if any). That logic is private at the moment but I'll work something out.

lwindolf: Thanks!

plugins/trayicon.py
```diff
@@ -174,10 +175,11 @@ def save_config(self, minimize_setting):
             f.write(minimize_setting)
 
     def trayicon_click(self, widget, data = None):
-        self.window.deiconify()
```
lwindolf, line 177: I fear we need to keep the deiconify() also. This is for window managers that still do icon minimization (think OpenStep, Enlightement) where you iconify Liferea and want it deiconified when clicking on the tray.

mhoran: I think that should be covered here: https://github.com/lwindolf/liferea/pull/826/commits/a1dda20aa7dacb429aa931d9b67cb475805bc3ab#diff-7386da2621ccaf16c69c72a556b8263eR1384.

lwindolf: Oh... I was to focussed on the trayicon code!

PR #844, opened 2020-06-17
Update metainfo path

nE0sIghT:
`/usr/share/appdata` is deprecated now:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,7 +17,7 @@ net.sourceforge.liferea.service: Makefile
 		mv $@.tmp $@
 
 
-appdatadir = $(datadir)/appdata
+appdatadir = $(datadir)/metainfo
 appdata_in_files = $(PACKAGE_TARNAME).appdata.xml.in
 appdata_DATA = $(appdata_in_files:.xml.in=.xml)
 


lwindolf:
@rich-coe @paulgevers Maybe you can advise on wether it is already time to do this switch from a distro perspective?

paulgevers:
@lwindolf I'm not totally sure if it matters for Debian. If I skim https://wiki.debian.org/AppStream/Guidelines and https://appstream.debian.org/ quickly, it seems that we pick up all the files:

> In Debian, we parse all XML provided by upstream projects as well as other metadata (.desktop-files, ...), and compile a single YAML metadata file from it, which is then shipped to users via APT.

rich-coe:
@lwindolf I'm not packaging liferea for any distro.  I don't have an opinion about this topic.

lwindolf:
@paulgevers Thanks for the info.

@rich-coe Sorry, totally mistook you for a maintainer :-)

lwindolf:
@nE0sIghT Thanks for the contribution! Merging this for 1.13.3

Issue #857, opened 2020-07-31
Build instructions missing a few items


BillDietrich:
On Ubuntu GNOME 20.04 at least, the source build instructions in the README are missing a few items: dependencies libtool and intltool, and a "sudo" on "make install".

I would like to fix this issue myself, as practice for submitting some source code changes to the project.  So please don't fix it yourself.

I have used GitHub only for single-person projects without branching.  Is the correct process to create this issue, create a branch, commit changes to the branch, then create a pull request (referencing the open issue) for the branch ?  Then you will evaluate the PR, and if approved, you will merge the branch into the master ?  Thanks.

BillDietrich:
I get an error when trying to push to the branch I created:

```
$ git status
On branch issue857
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	modified:   README.md

~/projects/liferea $ git diff
~/projects/liferea $ git commit -a -m 'in build instructions, add dependencies libtool and intltool, and "sudo" on "make install"'
[issue857 4b36e458] in build instructions, add dependencies libtool and intltool, and "sudo" on "make install"
 1 file changed, 4 insertions(+), 4 deletions(-)
~/projects/liferea $ git diff
~/projects/liferea $ git push -u origin issue857
Username for 'https://github.com': BillDietrich
Password for 'https://BillDietrich@github.com': 
remote: Permission to lwindolf/liferea.git denied to BillDietrich.
fatal: unable to access 'https://github.com/lwindolf/liferea.git/': The requested URL returned error: 403
~/projects/liferea $ git status
On branch issue857
nothing to commit, working tree clean
```

Leiaz:
That is because you don't have the right to push to this repository. In order to post a pull request, you must fork it on GitHub, to make your own Liferea repo on GitHub, then you push your branch with your changes on your Liferea repo on GitHub. After that, you can open a pull request and choose your branch. Github will even show a notification to directly open a pull request for that branch when you visit the Liferea repo.

BillDietrich:
Okay, I will do that, thanks.  But that is not the normal method, is it ?  I thought normally people could create a branch without forking a whole repo.  Maybe I'm wrong.

Leiaz:
No, you can only create branches if you have write access to the repo. Making a fork is the normal method on Github, unless you have write access to the repo.

BillDietrich:
Okay, let me see if I have this straight:
1- I fork the whole Liferea repo to make my own repo
[...]

PR #858, opened 2020-08-06
in build instructions, add dependencies libtool and intltool, and "sudo" on "make install"

BillDietrich:
Needed on Ubuntu 20.04 at least.


Leiaz:
@lwindolf On Debian testing too, libtool and intltool don't seem to be part of build-essential.

BillDietrich:
Not sure how I should cross-link these things.  Here is link to the original issue: https://github.com/lwindolf/liferea/issues/857

PR #861, opened 2020-08-18
Adding missing strings to the spanish translation

vosian:
Some missing strings were added tot he spanish translation

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -2344,3 +2344,2353 @@ msgstr "En descarga"
 
 #~ msgid "_Open link in:"
 #~ msgstr "Abrir enlace en:"
+# translation of es.po to Spanish
+# translation of es_099.po to Spanish
+# translation of es_094.po to Spanish
+# Traducción por Carlos Fenollosa <topopardo@topopardo.com>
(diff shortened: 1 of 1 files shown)


vosian:
Although I added some missing strings, there are a few that confuse me, what's the use of the strings that start with "_" like `_Copy Image Location` ? and then strings like `Open Link In _Tab`?

Leiaz:
They are [mnemonics](https://developer.gnome.org/gtk3/stable/GtkLabel.html). They are used for navigating the UI with the keyboard. The letter after the underscore in the code will be highlighted/underlined in the UI if you press the Alt key. Then pressing the letter moves the focus or select the associated widget/menu element.

vosian:
Indeed you are right, I have to fix that. I also noticed that strings with warnings on poedit don't show in liferea, most of those are because the original have exclamation marks but not the translation, but there are warnings over strings ending with ellipsis I don't know how to handle, it says strings have to end with a period

lwindolf:
Argh... auto-merge failed by reintroducing the duplication. Will manually merge to master.

PR #871, opened 2020-08-30
Javascript disable fix

lwindolf:
Merging missing code that was intended to rework the LifereaHtmlView to distinguish between internal and external web content loading.

Sorry about the other stuff that's in here too :-(


src/favicon.c
```diff
@@ -172,7 +172,7 @@ favicon_get_urls (subscriptionPtr subscription, const gchar *html_url)
 	}
 
 	/* case 2: */
-	if (html_url) {
+	if (html_url && strstr (html_url, "://")) {
 		tmp = g_strstrip (g_strdup (html_url));
```
Leiaz, line 176: It might be better to use Glib's String utility functions where possible.

src/parsers/rss_channel.c
```diff
@@ -353,14 +352,12 @@ rss_init_feed_handler (void)
...
 	}
-	
+
 	if (!rss_nstable) {
 		rss_nstable = g_hash_table_new (g_str_hash, g_str_equal);
 		ns_rss_ns_uri_table = g_hash_table_new (g_str_hash, g_str_equal);
 
-		/* register name space handlers */		
-		rss_add_ns_handler (ns_bC_get_handler ());
-		rss_add_ns_handler (ns_dc_get_handler ());
```
Leiaz, line 363: You removed blogChannel but ns_dc_get_handler() is removed too, was that intended ?

Leiaz (changes requested):
I have not looked at everything yet but the new feed auto discovery seems to be buggy. If I try to add urls that are not direct feed urls, I get "double free or corruption" crashes or garbage as source. For instance with https://www.mozilla.org I get gibberish as local file source (weird unicode characters, so likely reading memory it shouldn't).

```
double free or corruption (out)
--Type <RET> for more, q to quit, c to continue without paging--

Thread 1 "liferea" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff3a6c537 in __GI_abort () at abort.c:79
#2  0x00007ffff3ac5828 in __libc_message
    (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff3bd3e31 "%s\n")
    at ../sysdeps/posix/libc_fatal.c:155
#3  0x00007ffff3accb1a in malloc_printerr
    (str=str@entry=0x7ffff3bd6210 "double free or corruption (out)")
    at malloc.c:5347
#4  0x00007ffff3ace148 in _int_free
    (av=0x7ffff3c05b80 <main_arena>, p=0x7fffffffe180, have_lock=<optimized out>) at malloc.c:4314
#5  0x0000555555581657 in html_auto_discover_feed
    (data=0x555556e3c390 "\n\n\n<!DOCTYPE html>\n<html lang=\"en-US\" dir=\"ltr\" class=\"no-js\">\n<head prefix=\"og: http://ogp.me/ns#\">\n  <meta charset=\"utf-8\">\n  <meta http-equiv=\"X-UA-Compatible\" content=\"IE=Edge\">\n  <script>(functio"..., defaultBaseUri=0x555556f06cb0 "https://developer.mozilla.org/")
    at ../html.c:257
#6  0x000055555557f116 in feed_parser_auto_discover (ctxt=0x555556cd9190)
    at ../feed_parser.c:121
#7  0x000055555557f560 in feed_parse (ctxt=0x555556cd9190)
    at ../feed_parser.c:234
#8  0x000055555557eabf in feed_process_update_result
    (subscription=0x555556cdaa40, result=0x555556cc8da0, flags=9)
--Type <RET> for more, q to quit, c to continue without paging--
    at ../feed.c:318
#9  0x000055555558e60a in subscription_process_update_result
    (result=0x555556cc8da0, user_data=0x555556cdaa40, flags=9)
    at ../subscription.c:219
#10 0x0000555555590e52 in update_process_result_idle_cb
    (user_data=0x555556cd2bd0) at ../update.c:601
#11 0x00007ffff3dc24ce in g_main_context_dispatch ()
    at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#12 0x00007ffff3dc2880 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#13 0x00007ffff3dc290f in g_main_context_iteration ()
    at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#14 0x00007ffff400d02d in g_application_run ()
    at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#15 0x0000555555586ddc in liferea_application_new (argc=1, argv=0x7fffffffe5a8)
    at liferea_application.c:351
#16 0x000055555559116a in main (argc=1, argv=0x7fffffffe5a8) at main.c:54
(gdb) f 5
#5  0x0000555555581657 in html_auto_discover_feed (
    data=0x555556e3c390 "\n\n\n<!DOCTYPE html>\n<html lang=\"en-US\" dir=\"ltr\" class=\"no-js\">\n<head prefix=\"og: http://ogp.me/ns#\">\n  <meta charset=\"utf-8\">\n  <meta http-equiv=\"X-UA-Compatible\" content=\"IE=Edge\">\n  <script>(functio"..., defaultBaseUri=0x555556f06cb0 "https://developer.mozilla.org/")
    at ../html.c:257
257		g_free (baseUri);
```

src/html.c
```diff
@@ -206,28 +206,57 @@ search_links (const gchar* data, gint linkType)
...
+}
 
-	baseU = search_tag_link(data, "base", NULL);
-	if (!baseU)
-		baseU = baseUri;
+GSList *
+html_auto_discover_feed (const gchar* data, const gchar *defaultBaseUri)
+{
+	GSList		*iter, *links = NULL;
+	gchar		*baseUri;
```
Leiaz, line 222: Found a source of crashes/garbage source : this should be initialized to NULL or it will contain garbage and not be null in the if below.

lwindolf: Thanks for spotting this! It is amazing how useless Githubs CodeQL scanner is to not find this uninitialized variable. Maybe I misconfigured it, but so far it found nothing of relevance.

src/parsers/rss_channel.c
lwindolf: Oh... Reverted.

src/favicon.c
lwindolf: Good point. Changed.

Leiaz (approved):
Looks good :+1: 
I had a weird bug that disappeared after a `make clean`. The source seemed to be blank.
`** (liferea:349972): WARNING **: 22:28:33.934: The request for  could not be parsed!`
There may be an issue with the Makefile ...

lwindolf:
Thanks for retesting!

PR #875, opened 2020-09-05
Zoom : going back to normal size

Leiaz:
Because of the way zoom works in Liferea it is not possible to go back by zooming out after zooming in, the resulting size is slightly different which this is annoying. In Firefox for example, the zoom works differently, there seems to be a fixed list of zoom levels between 30% and 300%.
Here I just add an action to set the zoom back to 1:1 to the existing zoom in and out.

In the second commit I also remove what I think is obsolete code to make shortcuts for zoom works, but I'm not sure about that part. I added the shortcuts to the actions and it seems to work, and that way they are not completely hardcoded : they can be changed by plugins, etc ... But maybe there was another reason for those shortcuts to be handled in this way ?

### DIFF
--- a/glade/liferea_menu.ui
+++ b/glade/liferea_menu.ui
@@ -140,13 +140,17 @@
       <section>
         <item>
           <attribute name="action">app.zoom-in</attribute>
-          <attribute name="label" translatable="yes">_Increase Text Size</attribute>
+          <attribute name="label" translatable="yes">Zoom _In</attribute>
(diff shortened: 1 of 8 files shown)


lwindolf:
Nice! The extra key-handling was leftover code from the gtkmozembed times. Making hotkeys work inside Mozilla HTML widget was not so easy as it is in the well-behaved WebkitGtk widget :-)

PR #878, opened 2020-09-12
Set the urlentry progress bar, fixes #841

### LINKED ISSUE #841: Use a progressbar to indicate loading progress for built-in browser when previewing feed items
Liferea has a per-feed "auto-load item link in configured browser when selecting articles" setting that can be very useful when subscribed to items listings such as on Kijiji, to be able to quickly weed through stuff that is no longer available/listed.

The problem is that it sometimes takes 3-5 seconds for the Kijiji website to respond / for the page to load, even with Javascript disabled, and the pages are very similar to each other, so you sometimes think the page loaded instantly but it just hadn't loaded yet, because liferea doesn't "clear" the browser view on feed item change.

Ideally it should reset the webkit view so that it is a blank page before loading, but also, it should make use of the GtkEntry widget used for the addressbar for that built-in browser and show loading progress with it, like Epiphany does. Specifically, GtkEntry has a native method for showing a progressbar in it, [gtk_entry_set_progress_fraction()](https://developer.gnome.org/gtk3/stable/GtkEntry.html#gtk-entry-set-progress-fraction)

Leiaz:
Pass the WebKitWebView estimated progress value to the GtkEntry built-in progress bar.
And reset the bar to 0 when progress reaches 1, so the bar is only shown when loading.
This fixes #841.

### DIFF
--- a/src/ui/liferea_htmlview.c
+++ b/src/ui/liferea_htmlview.c
@@ -395,6 +395,13 @@ liferea_htmlview_title_changed (LifereaHtmlView *htmlview, const gchar *title)
 }
 
 void
+liferea_htmlview_progress_changed (LifereaHtmlView *htmlview, gdouble progress)
+{
+	double bar_progress = (progress == 1.0)?0.0:progress;
+	gtk_entry_set_progress_fraction (GTK_ENTRY (htmlview->urlentry), bar_progress);
+}
+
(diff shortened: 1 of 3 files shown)


lwindolf:
@Leiaz awesome works perfect

PR #882, opened 2020-09-18
Add the option to enable intelligent tracking prevention

### LINKED ISSUE #881: Implement enabling of Webkits Intelligent Tracking Protection
As done in Epiphany https://blogs.gnome.org/mcatanzaro/2020/09/16/epiphany-3-38-and-webkitgtk-2-30/ we should also enable Webkits ITP to prevent user tracking on internal web browser.

There should be an ITP preference in the privacy section. Default should be on.

Leiaz:
Adds the option to enable WebKit's intelligent tracking protection,
enabled by default, to the settings.
Adds a check button in the privacy tab, with a notice if WebKitGtk
version is too low.
Fixes #881 

For some reason the gschema file has dos line endings ... Should that be changed ?

Also, about the preference dialog, I don't see a way to add links or have wrapping text in the GtkCheckButton embedded label. I tried putting no text in that label, and all the text in a GtkLabel to the right, but then I could not have the mnemonic for accessibility work as it should, even after setting this label as labelling the checkbox. And also the text was not aligned with the text of the other GtkCheckButton above ... I wanted to include the link to https://webkit.org/tracking-prevention/ (or another ?) to have a more detailed explanation of the feature. The result is not really pretty ... If the label with the link is under the checkbox, it is also hard to align in a "nice" way ...

### DIFF
--- a/glade/prefs.ui
+++ b/glade/prefs.ui
@@ -1254,6 +1254,115 @@
                     <property name="top_attach">1</property>
                   </packing>
                 </child>
+                <child>
+                  <object class="GtkBox" id="itpbox">
+                    <property name="visible">True</property>
+                    <property name="can_focus">False</property>
+                    <child>
+                      <object class="GtkCheckButton" id="itpbtn">
+                        <property name="label" translatable="yes">_Intelligent Tracking Prevention. </property>
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="receives_default">False</property>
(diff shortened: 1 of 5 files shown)


mikelolasagasti:
> For some reason the gschema file has dos line endings ... Should that be changed ?

That may be my fault, no especial reason other than whatever happened in 2012 between me, SF and the patch I submitted. Please do.

lwindolf:
@Leiaz thanks for this!

Issue #884, opened 2020-09-22
Content of wrong feed is shown


Nudin:
From time to time the content of a news entry from an entirely different is shown instead of the correct one. See below for an example: The title and author if from a feed of the commits of a git repo, while the content is from a completely different feed (it's in german)…

![Bildschirmfoto](https://user-images.githubusercontent.com/509205/93937644-68621900-fd28-11ea-9030-dfb7b2aecd83.png)

Nudin:
Also: sometimes the audio-recording attached to one news item is shown on anther one (again different feed).

rich-coe:
I not sure about the feed, but on git, it will sometimes show ''wrong'' information, like the author/comment of the last commit on a branch when a new branch is created.

Nudin:
> I not sure about the feed, but on git, it will sometimes show ''wrong'' information, like the author/comment of the last commit on a branch when a new branch is created.

It's not a bug in the feed. The content shown isn't from github and it occurs with all kinds of feeds from blogs, newpages, etc.

rich-coe:
Your going to have to point to a public feed that exhibits the problem, that is reproducible. 
What I'm trying to explain is that complaining about the content of feed coming from a git repo is not conclusive evidence of a problem.

Leiaz:
@rich-coe That is an actual bug. It happened to me, but only once, on one item only. As I run master, plus sometimes other changes or temporary tests, and I couldn't reproduce it, I didn't search further.

@Nudin Can you reproduce it reliably ? Which version of Liferea are you using ?

Nudin:
Yes it is reproducable in the sense that the item stays 'brocken' also after restating liferea. I'm using version 1.13.2. At least the variant of the bug with the audio file already happened several months ago.

Nudin:
I looked into the liferea.db:
- In the 'item' table, I could find the item by it's title and it's description-field contains the correct content.
- I could however not find where the (wrong) content is stored.

Leiaz:
You are using "extract full content" on the misplace article content too ?

I think I know the issue : the metadata table still contains the metadata of deleted items. At least that what seams to happen.
```
sqlite> select count(distinct metadata.item_id) from metadata left join items on metadata.item_id = items.item_id where items.item_id is null;
2458
```
Full content is in the metadata table.

Nudin:
> You are using "extract full content" on the misplace article content too ?

I'm using it on the feed that the content is from – but not on the feed the header is from / where the content is shown.

[...]

PR #885, opened 2020-10-03
Add cleanup of metadata rows

Leiaz:
This adds deleting the metadata rows that don't have a corresponding item in the DB cleanup stage.

This is related to #884 but I'm not sure it fixes it. If the leftover rows are caused by crashes, then is enough to solve it. Anyway, I think it is logical to clean that table too.

Maybe there should be a better debug log to keep track that there *were* anomalous rows that needed to be removed.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -565,6 +565,10 @@ db_init (void)
 	db_exec ("DELETE FROM subscription_metadata WHERE node_id NOT IN "
           	 "(SELECT node_id FROM node);");
 
+	debug0 (DEBUG_DB, "Removing metadata without item...\n");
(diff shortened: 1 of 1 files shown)


lwindolf:
This cleanup is a good idea and might cleanup a lot of peoples large DB files!

I think the actual root cause of #884 lies in this line from src/db.c
 
        sql = sqlite3_mprintf ("SELECT MAX(item_id) FROM items");

Which is a bad practice DB schema-wise. A better solution is to have an ever-increasing id (of course leading to overflow in the long term). Something like having a static "last_new_item_id" in src/db.c which is initialized on startup with said "SELECT MAX..." statement and only ever increased afterwards.

PR #886, opened 2020-10-03
Put comments in another div outside of content

### LINKED ISSUE #883: Feed with comments : last comment replaces all content
This may be due to Readability.js. The feed is displayed correctly with c680b8e204282236dc21e08624d8, I didn't bisect though. I have (external now) javascript disabled.

Using master, when viewing https://www.pine64.org/feed/ the content of the article which is usually a summary of the blog item is completely replaced by the latest comment.

Leiaz:
Fixes #883

Puts comments in another div outside of content to exclude it from Readability.js as suggested.

### DIFF
--- a/xslt/item.xml.in
+++ b/xslt/item.xml.in
@@ -301,8 +301,10 @@
     <!-- the real text content -->
     <xsl:value-of select="description" disable-output-escaping='yes'/>
   </p>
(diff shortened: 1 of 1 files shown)


lwindolf:
Thanks a lot!

PR #890, opened 2020-10-06
Update Indonesian translation

samsulmaarif:


lwindolf:
Thank you for this localization update!

PR #891, opened 2020-10-06
Update bookmark sites

mikelolasagasti:
- Switch blogmarks to https
- Remove buddymarks: no DNS entries
- Remove citeulike.org: ceased operations on 2019-03-30
- Remove Google Plus: ceased operations on 2019-04-02
- Change default from now defunct del.icio.us to Pocket

### DIFF
--- a/src/social.c
+++ b/src/social.c
@@ -101,14 +101,11 @@ social_init (void)
 {
 	gchar *tmp;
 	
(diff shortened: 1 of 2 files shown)


src/social.c
```diff
@@ -123,7 +120,7 @@ social_init (void)
 	g_free (tmp);
 	
 	if (!bookmarkSite)
-		social_set_bookmark_site ("del.icio.us");		/* set default if necessary */
```
lwindolf, line 126: In a way I'm really sad the del.icio.us is gone. It was nice and simple.

PR #893, opened 2020-10-08
Add Persian opml localization.

KevinScruff:


lwindolf:
Thank you for this localisation!

PR #894, opened 2020-10-08
Update French translation

yannicka:


lwindolf:
Thank you for the update!

PR #912, opened 2020-11-20
Allow multiple feeds in the same notification

tatokis:
This PR modifies the libnotify plugin to include multiple feed updates in the same notification if they are updated within a certain time period.
This way, the user does not get spammed with notifications if they have multiple feed updates at once.

### DIFF
--- a/plugins/libnotify.py
+++ b/plugins/libnotify.py
@@ -2,6 +2,7 @@
 # libnotify Popup Notifications Plugin
 #
 # Copyright (C) 2013-2015 Lars Windolf <lars.windolf@gmx.de>
+# Copyright (C) 2020 Tasos Sahanidis <tasos@tasossah.com>
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Library General Public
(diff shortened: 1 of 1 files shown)


lwindolf:
The merging of new subscription names does work. But there is an issue when using GNOME Shell that `<br>` is shown as text and not causing a line break.

tatokis:
I updated it and added a newline instead. Does that fix the issue on your end?

lwindolf:
Nope, not really in GNOME Shell it seems newlines are also ignored. It shows the feed names separated by a space, which is very confusing.

lwindolf:
Argh... actually clicked "comment and merge" instead of just commenting.

tatokis:
> I believe the cause to be https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1118

Ugh.

I personally don't really like the commas, as it makes it difficult to tell where the feed names end. Some feeds even have commas in their titles too...

3bdad610e6c8
2f9098a18687
5ac58bca27ff

I think it should be made configurable with a LineEdit that allows the user to enter their own separator with the comma being the default.
Alternatively, we could check for gnome and use commas only there, but that's a bit messy. That said, I haven't seen any other desktop environment disallow newlines.
What do you think?

lwindolf:
You are right even a comma is not safe. I just discovered that just the short view is without line breaks and  the detail view looks perfectly fine. I wouldn't do any extra solution for GNOME too. So let's go with your solution!

lwindolf:
I've merged your solution now again (by reverting the revert). I hope you do not mind the possible missing git commit attribution. Sorry about the chaos!

Thank you a lot for contributing this! Will be released with 1.13.4

tatokis:
No worries. Glad to also hear it only applies on the preview/short view as well. Makes things easier.

> I hope you do not mind the possible missing git commit attribution. 

It's okay, although technically the attribution is still there due to the revert

Thanks!

PR #918, opened 2020-12-06
Implement support for subscribing to LD+Json metadata listings

lwindolf:
Here are some example subscription sources:

- https://www.ticketmaster.com/discover/concerts?geoHash=dr5r&radius=100&sort=date%2Casc&unit=miles&daterange=all
- https://www.songkick.com/metro-areas/28443-germany-berlin


src/metadata.c
```diff
@@ -91,9 +91,6 @@ metadata_init (void)
 	metadata_type_register ("source",		METADATA_TYPE_URL);
 	metadata_type_register ("coverage",		METADATA_TYPE_HTML);
 
-	/* types for photo blogs */
```
lwindolf, line 94: Was found to be unused.

src/parsers/ldjson_feed.c
```diff
@@ -0,0 +1,364 @@
...
+	if (eventStatus && eventStatus == g_strstr_len (eventStatus, -1, "https://schema.org/Event"))
+		eventStatus += 24;
+	else
+		eventStatus = NULL;
+
+	if (tmp = json_get_string (node, "startDate")) {
+		// schema.org says startDate should be ISO8601, but RFC822
+		// is seen to often. So fuzzy match date format.
+		if (strstr (tmp, " "))
+			ctxt->item->time = date_parse_RFC822 (g_strdup (tmp));
```
Leiaz, line 134: Sorry for being late, I haven't read everything, but there is a leak here. The `g_strdup` are not necessary.

Also the examples display `There were errors while parsing this feed!` for me. (But the items appears to work anyway)

lwindolf: Thank you for taking the time to review!!! I'll fix the g_strdup().

About the error message: I'm working on a refactoring of the error handling, so this will be addressed soon.

PR #919, opened 2020-12-16
Fixes #903: Fix autodiscovery endless loop

### LINKED ISSUE #903: Feed detection and redirection endless loop
When trying to find a feed url, Liferea can get into an endless loop, and the UI doesn't really show what is happening. The main hint is that Liferea is using a lot of bandwith while not showing anything in the new feed.
Here is an example of a problematic site : https://www.groupe-ldlc.com/communiques-de-presse/
The page advertises with `<link rel="alternate" type="application/rss+xml" />` a feed url that doesn't really exist but instead of throwing a 404 redirect to https://www.groupe-ldlc.com/, which advertises a feed that isn't there and redirects ...
In fact, there are three feed urls given and only the last one is an actual feed.

So the fault is partly theirs for giving bad links, but it would be nice if Liferea didn't fall into a loop. There shouldn't be a big number of attempts before finding the feed in normal cases, so perhaps we could count those and limit to a fixed number like 3 or 4 ...

lwindolf:
- Add audodiscovery redirect counter to subscription
- Stops after 5 redirects
- Drop mutation of ctxt->failed use return codes instead

### DIFF
--- a/src/feed_parser.c
+++ b/src/feed_parser.c
@@ -93,7 +93,7 @@ feed_create_parser_ctxt (void)
 
 void
 feed_free_parser_ctxt (feedParserCtxtPtr ctxt)
-{https://www.groupe-ldlc.com/communiques-de-presse/
+{
 	if (ctxt) {
 		/* Don't free the itemset! */
 		g_hash_table_destroy (ctxt->tmpdata);
--- a/src/feed_parser.c
+++ b/src/feed_parser.c
@@ -178,7 +178,6 @@ feed_parse (feedParserCtxtPtr ctxt)
 	xmlNodePtr	xmlNode = NULL, htmlNode = NULL;
 	xmlDocPtr	xmlDoc, htmlDoc;
 	gboolean	autoDiscovery = FALSE, success = FALSE;
-	gboolean	htmlParsing = FALSE;
 
 	debug_enter ("feed_parse");
 
--- a/src/feed_parser.c
+++ b/src/feed_parser.c
@@ -30,6 +30,8 @@
 #include "parsers/ldjson_feed.h"
 #include "parsers/rss_channel.h"
 
+#define AUTO_DISCOVERY_MAX_REDIRECTS	5
+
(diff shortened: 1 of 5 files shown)


src/feed_parser.c
```diff
@@ -128,18 +130,20 @@ feed_parser_auto_discover (feedParserCtxtPtr ctxt)
...
-		debug0 (DEBUG_UPDATE, "No feed link found!");
-		g_string_append (ctxt->feed->parseErrors, _("The URL you want Liferea to subscribe to points to a webpage and the auto discovery found no feeds on this page. Maybe this webpage just does not support feed auto discovery."));
+
+		return TRUE;
 	}
+
+	debug0 (DEBUG_UPDATE, "No feed link found!");
+	g_string_append (ctxt->feed->parseErrors, _("The URL you want Liferea to subscribe to points to a webpage and the auto discovery found no feeds on this page. Maybe this webpage just does not support feed auto discovery."));
+	return FALSE;
 }
```
Leiaz, line 147: I had not noticed the implicit return here before, in the sea of mostly useless warnings. Hopefully the compiler would have complained louder if we had tried to use the value.

src/feed_parser.c
```diff
@@ -174,15 +177,13 @@ feed_parse (feedParserCtxtPtr ctxt)
 {
 	xmlNodePtr	xmlNode = NULL, htmlNode = NULL;
 	xmlDocPtr	xmlDoc, htmlDoc;
-	gboolean	success = FALSE;
+	gboolean	autoDiscovery = FALSE, success = FALSE;
 	gboolean	htmlParsing = FALSE;
```
Leiaz, line 181: On the topic of warnings, the `htmlParsing` variable is unused.

Leiaz (approved):

src/feed_parser.c
lwindolf: Fixed

src/feed_parser.c
lwindolf: True the warnings are a hassle. I'm trying to reduce them more and more.

Issue #332, opened 2016-04-07
Search is case sensitive for everything uppercase


hphde:
Hi,

I was searching a string and wrote the word lower case and it wasn't found. Obviously it was there and then I saw it was written all uppercase.

I tried some tests and it seems words starting with a single starting uppercase character are not affected.

Kind regards,
HP

nekohayo:
That's interesting, from my testing I was under the impression that everything was case sensitive all the time... and I would love to have a checkbox in the search conditions somewhere to be able to set a particular condition to be case (in)sensitive. Currently, I have to deduplicate 3-4 times my search folders to account for "most" possible scenarios, but I'd rather really have the ability to have case insensitive searches.

PR #920, opened 2020-12-16
Fixes #908,#332: Search folder rules are not case-insensitive

### LINKED ISSUE #908: Searches are case-sensitive, unlike what the documentation / user manual says
http://lzone.de/liferea/help110/searching_en.html says that everything in the table (such as the "item title does contain" rule) is case **in**sensitive, whereas I can reliably demonstrate that to be false. It has been case-sensitive for years, at least since 2016 if issue #332 is any indication.

Looking at the documentation tonight and reading that it is supposedly case-insensitive, I was very surprised, so I tested again tonight, and can confirm the app behavior contradicts the documentation. There is a difference in search results if I search for "marketing" vs "Marketing" in my feed item titles with a Liferea search folder. The results count, and the item results you can see with the naked eye, are completely different. For example:
* if I search for "arketing" I get 67 results;
* if I search for "marketing" I get only 27 results;
* if I search for "Marketing" I get 40 results.

40+27 = 67. Busted!

I'd like it all to be case-insensitive like the documentation states, or, even better, to have those conditions offer a case-insensitive vs case-sensitive variant of everything.

It would make my search queries much simpler and more reliable.

lwindolf:


src/rule.c
```diff
@@ -72,25 +72,56 @@ rule_new (const gchar *ruleId,
...
+	g_free (rule->valueCaseFolded);
 	g_free (rule);
 }
 
+/* case insensitive strcmp helper function
+
+   To avoid half of the g_utf8_casefold we expect the 2nd value to be already
+   case folded!
+ */
+const gchar *
```
Leiaz, line 100: This could be `static`.

Leiaz (approved):

src/rule.c
lwindolf: Is applied.

PR #928, opened 2020-12-27
Add gettext support for the plugin-installer.py and its Polish translation

sunwire:


lwindolf:
Thanks for adding translation to the plugin installer and updating the PO translation!

Happy New Year!

sunwire:
Thank You and Happy New Year.

PR #929, opened 2020-12-27
Update AppData with release dates

mikelolasagasti:
New release info is required to be added in appdata file so flatpak build reports correct version. 

I added this patch to a new flatpak build in flathub.


lwindolf:
Arg... I cannot promise not to forget that the release needs to be listed there :-(

PR #933, opened 2020-12-31
Add gettext support for the notify plugin

sunwire:


lwindolf:
Thanks for improving this!

PR #934, opened 2020-12-31
Add gettext support for the trayicon plugin

sunwire:


lwindolf:
Thanks for improving this!

PR #936, opened 2021-01-01
Some python fixes

sunwire:
I tested all the changes and the plugins work as before the changes.
Remove unused imports
Remove not needed return
Remove space between func name and (
small changes to satisfy pylint


lwindolf:
Great code cleanup! To ensure it stays clean we could add a pylint invocation in "make test" this way the Github Actions CI will catch stuff.

Issue #931, opened 2020-12-30
Headerbar plugin: bind F10 keyboard shortcut to toggling the main menubutton


nekohayo:
Most/all GNOME apps use the de-facto standard F10 keyboard shortcut for toggling the main menubutton, so if possible, this app should as well.

This is also documented as a standard shortcut in https://developer.gnome.org/hig/stable/keyboard-input.html.en

sunwire:
@nekohayo Can You test this PR https://github.com/lwindolf/liferea/pull/937 ?

PR #937, opened 2021-01-02
Bind the F10 key to the hamburger menu

sunwire:
This should solve https://github.com/lwindolf/liferea/issues/931
I've tested on mate-desktop and works fine but it needs testing on other DMs (Gnome3 , Xfce etc.)


lwindolf:
Works great on GNOME. Thanks!

Issue #867, opened 2020-08-29
problem with rss feed


cheese1:
hi

i get errors while reading:
https://security.netapp.com/advisory/index.xml

i get:
> There were errors while parsing this feed!
> Details
> Could not detect the type of this feed! Please check if the source really points to a resource provided in one of the supported syndication formats!
> 
> XML Parser Output:
> Source points to HTML document.

adding "xmllint --recover -" did not help
how would i debug that?

ps.: the feed is quite huge

cheese1:
Feeder under android parses that feed without problems

Leiaz:
Huge indeed. It is 22M, and it chokes the libxml parser with `internal error: Huge input lookup`. libxml as a `XML_PARSE_HUGE` option but I don't see how to pass it to the SAX parser ...

cheese1:
by reducing the file size i was able to read the feed.
i used a filter to remove empty lines and spaces/tabs at the start of lines
` sed -e 's/^\s*//' -e '/^$/d' `

seems to solve at least my problem

lwindolf:
Having this feed in my test case set I noticed it suddenly working. I think at latest after the Ubuntu 22.04 upgrade. So I guess some limits in libxml2 might have become bigger (or the feed smaller).

@cheese1 Closing this as it is an old ticket (and works now).

PR #938, opened 2021-01-02
Fixes #809: Improved error presentation for feed fetch errors

### LINKED ISSUE #809: Use a real infobar and allow showing details/logs for feed parsing errors
When a feed parsing error occurs, this yellow banner is shown:

However, as illustrated in #808, this isn't really helpful for users to either report bugs on Liferea or to notify the website owners. I mean sure, I know there were errors... but Liferea is not telling me what they are and why it can't cope with them.

Ideally, this yellow banner should be replaced by a standard GtkInfobar widget, with an action button to "Show details..." that could display the errors in a scrollable (and copy-able) textview in a separate window.

Ideally, that window should be [auto-sized](https://fortintam.com/blog/dont-make-me-scroll/) to avoid the need for scrolling, but anything is better than nothing for now.

Heck, it could simply dump a .txt (or .xml, or something with the proper file extension) error log file into /tmp/ (or something in the user XDG cache directory where liferea stores its cache) and open it with the user's default text editor using xdg-open and voilà, you get syntax highlighting for free, and no need to create an additional UI dialog in Liferea ;) it could even prepend the log with some notes about reporting it in Liferea or pre-written text to make it easy for users to include in an email to website owners with broken feeds (in the cases where it's really broken beyond any hope of Liferea being able to parse it).

lwindolf:
Todo:

- [x] HTTP 410 is properly indicated
- [x] HTTP 401 is properly indicated
- [x] Details for HTTP status code messages 
- [x] Details for parser error messages 
- [x] Details for filter error messages 
- [x] CSS looks ugly
- [x] HTTP 401 node needs to be unavailable
- [x] HTTP 410 node needs to be unavailable

Here is how it currently looks like

error-display


src/xml.c
```diff
@@ -525,22 +525,8 @@ xml_get_ns_attribute (xmlNodePtr node, const gchar *name, const gchar *namespace
 	return xmlGetNsProp (node, BAD_CAST name, BAD_CAST namespace);
 }
 
-static void
-liferea_xml_errorSAXFunc (void * ctx, const char * msg,...)
-{
```
Leiaz, line 530: Is this really a duplicate that returns the same errors ?

For the feed given in #867 it returned `internal error: Huge input lookup` but the new error reporting just reports feed discovery failed. This feed is a weird case though, as parsing doesn't fail because of bad xml ...

Leiaz (approved):
It looks really nice :+1:

src/xml.c
lwindolf: Thanks for the hint. I'll retest with this feed!

lwindolf: I'm readding the liferea_xml_errorSAXFunc() callback so we get extra errors on the console. But due to the signature being different from xml_buffer_parse_error() where an errorCtxtPtr is being passed (while liferea_xml_errorSAXFunc() gets a xmlSaxHandlerPtr) it is impossible to pass the error to the GUI.

Issue #950, opened 2021-01-14
license of plugins/getfocus.py is not properly documented


paulgevers:
The license statement as currently found in plugins/getfocus.py is not great. There are two issues.

1. The license statement says the license can be found in a file called COPYING.LIB, but it's not included.
2. The license statement says "GNU Library General Public License as published by the Free Software Foundation; either version 3 of the license, or (at your option) any later version". However the LGPL was renamed from "GNU Library General Public License" to "GNU Lesser General Public License" in version 3, so the claimed license version doesn't really exist.

Thanks for the work on liferea, I enjoy it every day.

sunwire:
The same license is in this files:
plugins/headerbar.py
plugins/gnome-keyring.py
plugins/libnotify.py
plugins/trayicon.py
plugins/media-player.py
src/ui/ui_dnd.h
src/ui/search_folder_dialog.h
src/ui/search_dialog.h
src/ui/liferea_shell.h
src/ui/rule_editor.h
src/ui/liferea_dialog.h
src/ui/item_list_view.h
src/parsers/ns_slash.h
src/parsers/ns_syn.h

paulgevers:
@sunwire you're right, it's not a unique problem, I spotted it this time because the file is new. However, COPYING.LIB is till missing.

Except for headerbar.py, all others have version 2, so they can't share the same COPYING.LIB file *and* problem 2 doesn't apply.

lwindolf:
These are mostly copy&paste errors :-)

The good thing I believe in all cases the author intention is clear and we can fix the reference to the correct license file.

paulgevers:
Thanks.

lwindolf:
Reopening to cover the other issues mentioned by @sunwire

sunwire:
@lwindolf Maybe instead of putting the content of the license in the code, put it in a separate file and leave only a mention of the license in the code e.g.:
License: GPLv3 see licenses/GPLv3.txt
License: LGPLv3 see licenses/LGPLv3.txt
License: MIT see licenses/MIT.txt
Etc.

lwindolf:
@sunwire Nope that would be a license violation for all those licenses.

lwindolf:
@paulgevers Please review the change for problems!

 @yaneti I found your bold-unread plugin was missing a license. I've added LGPLv3+ because I guess this was your intention. Please let me know if this is incorrect!

paulgevers:
@lwindolf looks good to me...

[...]

PR #951, opened 2021-01-14
Correcting the license

sunwire:
I hope this fix https://github.com/lwindolf/liferea/issues/950


lwindolf:
Thanks for fixing this!

PR #953, opened 2021-01-17
l10n: update French translation

guilieb:
Hi,

Here is an update for the French translation on the 1.12 branch.

Thanks,


lwindolf:
Thank you!

Issue #901, opened 2020-10-15
Layout not remembered 


GaryGate15:
New issue with latest version 1.13.3:

In Normal view, the position of divider between Headlines and article content is not remembered. After a restart, it is always at the bottom, with only a minimal slice of the article content shown.

Leiaz:
This is not exactly new, and it looks tricky to fix.

See #570 #590 #762 

To resume, we can keep this bug, or reopen bug #540 : articles or headlines being squeezed in the same way on window resize instead.

We are currently restoring the previous value at start for the pane's separator position, but as explained in #570, the value then jumps around.
This is a Gtk bug that was first reported in 2014 : https://bugzilla.gnome.org/show_bug.cgi?id=733638 and as noted in #590, also in their new Gitlab here : https://gitlab.gnome.org/GNOME/gtk/-/issues/1959

miwalter:
Liferea always resets window size to some default value. Doesn't remember window size or positions at all ;)

Wide view.

konomikitten:
This issue is persistent and annoying I find that every time I restart Liferea the preview bar slowly moves down the pane until it is no longer visible.

Distro: Debian GNU/Linux bullseye/sid x86_64
Liferea: 1.13.3

nekohayo:
It is strange that this did not happen in 1.12.x but happens in 1.13.x, given that both are using GTK3...

Looking in dconf-editor at the net.sf.liferea "last-hpane-pos" key value, that value is set/saved only when Liferea quits (not while you are dragging the widget, nor on startup). On startup, that value is not changed. So my question is... is Liferea telling GTK to set that value sometime too soon, before the entirety of the UI is set up? Would it change anything if it set that value at the beginning of the startup (before showing the widgets) and also later "at the last minute"?

...because if I open the GTK Inspector, find the GtkPaned widget (in this case named "normalViewPane" and set its "position" property to anything else (like "200"), the pane positions itself correctly. Even incrementing/decrementing by one pixel the position sets it accurately. So in the case at hand, I'm just guessing, but maybe it is primarily a startup timing issue somehow?

nekohayo:
Hi @lwindolf, I have narrowed down the issue today! This bug happens ONLY when starting Liferea in windowed mode instead of starting it maximized. Additionally, like we see in issue #909, it might get subtly influenced by whether or not the headerbar plugin is active.

[...]

PR #964, opened 2021-02-20
Pane position workaround

sunwire:
It is a workaround  for https://github.com/lwindolf/liferea/issues/901 it doesn't fix it but it works.
The issue #901 should remain open until properly resolved and then this plugin must be removed.
The plugin is enabled by default.


lwindolf:
@sunwire Thanks for providing this workaround!!!

PR #976, opened 2021-04-05
Register liferea custom scheme, fixes #973

### LINKED ISSUE #973: liferea.css is no longer applied after update webkit2gtk-2.30.6 to 2.32.0 in archlinux
Liferea 1.13.5 and also newest version from master (commit b3de8af3c8b1a04df33211053cba9b6a74b8a2de).

With webkit2gtk-2.30.6 everything is fine, with webkit2gtk-2.32.0 liferea.css doesn't work - see screenshots with activated webkit2-inspector. 

webkit2gtk-2 30 6
webkit2gtk-2 32 0

With webkit2gtk-2.32.0 I get the message: "User Agent Style Sheet cannot be modified." and as error-message: "Not allowed to load local resource: file:///home/harald/.cache/liferea/style.css?1617130761"
Perhaps of this?: The changelog (https://webkitgtk.org/2021/03/26/webkitgtk2.32.0-released.html) states: "New API to remove individual scripts/stylesheets using WebKitUserContentManager."

The only workaround for me is to downgrade to webkit2gtk-2.30.6.

Am I missing something?

Best regards
harald

Leiaz:
It used to be enough to declare the liferea custom scheme as local to
access resources with file scheme, but for WebKit2Gtk >= 2.32 it looks
like it is necessary to register the custom scheme with a handler.
This fixes #973

The handler doesn't do anything interesting for now as we just pass the
content with webkit_web_view_load_bytes and use the file scheme to
access resources, but it could be used to load Liferea resources in the
future ...

### DIFF
--- a/src/webkit/webkit.c
+++ b/src/webkit/webkit.c
@@ -392,12 +392,29 @@ liferea_webkit_impl_download_started (WebKitWebContext	*context,
 }
 
 static void
+liferea_webkit_handle_liferea_scheme (WebKitURISchemeRequest *request, gpointer user_data)
+{
+	const gchar *uri = webkit_uri_scheme_request_get_uri (request);
+	GInputStream *stream;
+	gssize length;
+	gchar *contents;
+
+	contents = g_strdup_printf ("Placeholder handler for liferea scheme. URI requested : %s", uri);
+	length = (gssize) strlen (contents);
+	stream = g_memory_input_stream_new_from_data (contents, length, g_free);
+	webkit_uri_scheme_request_finish (request, stream, length, "text/plain");
+	g_object_unref (stream);
+}
+
+static void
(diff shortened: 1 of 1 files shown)


lwindolf:
Thanks for fixing this!!!

mozbugbox:
I've just done something similar  locally the other day. My intent was to load the item content through a URI like `liferea:///item/182392`. 

That way, the item page will be add into the history so that I can use the back action to go back to the item page from a normal http page.

I have some forum like feed where a thread was shown as an item. When I clicked a link in the thread to load a post, I'd like to be able to go back to the item page for other post in the thread. Reload the item works but the item view always scrolled to the top when reload. With back action, the page would scroll to where it left.

[liferea-scheme.txt](https://github.com/lwindolf/liferea/files/6275852/liferea-scheme.txt)

PR #980, opened 2021-04-17
Update for sq translation

ujdhesa:
Here you are an update for sq translation.

TIA

Besnik


lwindolf:
@ujdhesa I fear your translation is build on a very old (1.11) version of the code. Your fork is >1000 commits behind. It does not really match against recent 1.13 code.

lwindolf:
@ujdhesa Thanks for updating the sq translation once more. Your work is really appreciated!!!

PR #984, opened 2021-04-30
Replaces GtkFileChooser with native, fixes #983

### LINKED ISSUE #983: Use GtkFileChooserNative
Liferea doesn't seem to use [GtkFileChooserNative](https://developer.gnome.org/gtk3/stable/gtk3-GtkFileChooserNative.html) that supports [portal API](https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-org.freedesktop.portal.FileChooser) but [GtkFileChooser](https://developer.gnome.org/gtk3/stable/GtkFileChooser.html). 

Portals allow to dynamically grant access to users rather than [hard-coding arbitrary filesystem access](https://github.com/flathub/net.sourceforge.liferea/blob/3ab07232cc567f2a8e2e283ca74f2c4e562bd388/net.sourceforge.liferea.json#L23) for example for an application. The primary benefit here is tighter sandboxing; also better desktop integration (*) and both [Snap](https://snapcraft.io/docs/xdg-desktop-portals) and [Flatpak](https://flatpak.github.io/xdg-desktop-portal/portal-docs.html) support them.

(*) When an application is running on a desktop environment which does not use GTK+, the desktop environment could set up things to use its native open/save dialogs.

Leiaz:
@bbhtt Please test that it does what you want.

### DIFF
--- a/src/ui/ui_common.c
+++ b/src/ui/ui_common.c
@@ -123,13 +123,12 @@ ui_show_info_box (const char *format, ...)
 }
 
 struct file_chooser_tuple {
-	GtkWidget *dialog;
 	fileChoosenCallback func;
 	gpointer user_data;
 };
 
 static void
-ui_choose_file_save_cb (GtkDialog *dialog, gint response_id, gpointer user_data)
+ui_choose_file_save_cb (GtkNativeDialog *dialog, gint response_id, gpointer user_data)
 {
 	struct file_chooser_tuple *tuple = (struct file_chooser_tuple*)user_data;
 	gchar *filename;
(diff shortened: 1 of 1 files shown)


bbhtt:
Thanks! 

Seems to now use the Filechooser portal for me:

```
method call time=1619836905.377955 sender=:1.59 -> destination=org.freedesktop.portal.Desktop serial=72 path=/org/freedesktop/portal/desktop; interface=org.freedesktop.portal.FileChooser; member=OpenFile
   string "x11:4600007"
   string "Import Feed List"
   array [
      dict entry(
         string "handle_token"
         variant             string "gtk1746656723"
      )
      dict entry(
         string "multiple"
         variant             boolean false
      )
      dict entry(
         string "directory"
         variant             boolean false
      )
      dict entry(
         string "accept_label"
         variant             string "Import"
      )
      dict entry(
         string "modal"
         variant             boolean true
      )
      dict entry(
         string "filters"
         variant             array [
               struct {
                  string "OPML Files"
                  array [
                     struct {
                        uint32 0
                        string "*.opml"
                     }
                     struct {
                        uint32 0
                        string "*.xml"
                     }
                  ]
               }
               struct {
                  string "All Files"
                  array [
                     struct {
                        uint32 0
                        string "*"
                     }
                  ]
               }
            ]
      )
      dict entry(
         string "current_file"
         variant             array of bytes "/home/user" + \0
      )
   ]
```
cc @mikelolasagasti

PR #993, opened 2021-05-25
net: Do not add a second content decoder to the libsoup session

yaneti:
The default soup session constructor already adds a content decoder.
This was harmless so far as long as there was only the deflate/gzip decoder,
which didn't mind being passed plain text content that was already
ungzipped, but becomes more problematic now that libsoup also supports
brotli compression.

See https://gitlab.gnome.org/GNOME/libsoup/-/issues/146 for some context


lwindolf:
Thank you for debugging this and contributing the bugfix!

PR #995, opened 2021-05-31
Update cs.po

RadimNo:


lwindolf:
Thank you for updating the Czech translation!

Issue #788, opened 2020-03-08
RFC: feature request: deferred marking of read items 


rich-coe:
I am currently using a feature where in the Unread list, after reading an item, it does not immediately delete the item from the unread list when moving to the next item, but marks the item as 'deferred read' as a transient state.   This allows me to go back to a previously read item if I mistakenly skip one (fat finger a key), or want to go back a re-read a previously read item in more detail.  I currently flush the deferred read items by switching to another item feed, but there could be a menu item that 'executes' this action.

Please let me know if you would find this feature useful.
Would this feature need to be optional to keep the current behaviour or should it always be there?

rich-coe:
Note this is a request for comments in preparation for a pull request to @lwidolf.

lwindolf:
Do you use the "Previous Item" feature to go back to the previous item?

This feature is currently only accessible via mouse or menu.

The probably should be a corresponding hotkey like Ctrl-P that matches Ctrl-N for "Next headline".

What do you think about a hotkey solution to this problem?

lwindolf:
@rich-coe ping

rich-coe:
@lwindolf  I'll make a PR this week.

lwindolf:
Awesome!

rich-coe:
Add UI preference 'Defer removing read items from Search Folder'.
Don't delete items from folder, like 'Unread', until leaving the folder.

Uses transient 'isHidden' flag on the item to set the font.

When reading items from a Search Folder, like 'Unread', defer removing the items from the UI until you leave the folder.  This allows re-reading read items or marking them unread.  Read items appear in light-weight font.

lwindolf:
Added with merge of #997

PR #997, opened 2021-06-02
changes for #788 - introduce defer delete mode for Search folders

rich-coe:
Add UI preference 'Defer removing read items from Search Folder'.
Don't delete items from folder, like 'Unread', until leaving the folder.

### DIFF
--- a/doc/html/preferences_en.html
+++ b/doc/html/preferences_en.html
@@ -92,6 +92,12 @@
 	<b>Default View Mode</b> Select your favourite viewing mode here.
 	</li>
 
+        <li>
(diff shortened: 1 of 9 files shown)


lwindolf:
Sorry for not yet having tested this. Will look into it for 1.13.8

lwindolf:
@rich-coe with 41c7eb800eacf3dfeaa3d81d3be2b515b25c4a5b I rephrase the option to reflect that it also affects folders. I also changed "Search Folders" to be lower case to match other preference labels.

I hope you do not mind.

Many thanks for the feature which works well and sorry for the long review wait!

Keep up the good work :-)

Issue #998, opened 2021-06-02
Option to display fixed width date/time for headlines


rich-coe:
Add a headline option to always display date/time as %m/%d %H:%M for current year, and %Y/%m/%d for previous years.
Leave as default Today, Yesterday, day names for last week.

rich-coe:
Add user preference to display date/time in fixed format.
This year displayed as %m/%d %H:%M, otherwise display as %Y/%m/%d.

bhoefer2015:
> Leave as default Today, Yesterday, day names for last week.

That is actually a thing about Liferea that annoys me -- replacing the date column's "%b %d" format with "%a" (or "Today" or "Yesterday") for the past 7 days.  I'd rather it always show the name (or number) of the month with the day of the month, at least for the current year.

Although it would be crowded, "%a %b %d" format would also be acceptable.

(
%a     locale's abbreviated weekday name (e.g., Sun)
%b     locale's abbreviated month name (e.g., Jan)
%d     day of month (e.g., 01)
)

lwindolf:
I really understand your wishes, but I don't want to maintain and support multiple date format variants. I want to ensure that 1 way works 100%

PR #999, opened 2021-06-02
feature #998 user pref for fixed width date in headline

rich-coe:
Add user preference to use fixed width date in headline.
Dates in this year are displayed %m/%d %H:%M, otherwise they are displayed %Y/%m/%d.


src/item.c
```diff
@@ -48,7 +48,12 @@ item_new (void)
 itemPtr
 item_load (gulong id)
 {
-	return db_item_load (id);
```
lwindolf, line 51: While the modification of the item loading interface is functionally valid I still think we should do this UI thingy only in the UI code in item_list_view_update_item_internal(). I think moving it there might even save as from a lot of extra string formattings that do never touch the UI (think search folder rebuilds, our item list merging on feed updates).

src/ui/item_list_view.c
```diff
@@ -471,13 +471,19 @@ item_list_view_update_item_internal (ItemListView *ilv, itemPtr item, GtkTreeIte
 	gchar		*title, *time_str;
 	const GIcon	*state_icon;
 	gint		state = 0;
+        gboolean        fixeddatefmt;
```
lwindolf, line 474: Regarding indentation: you do not need to fix it... But let me ask which IDE do you use? Would having VIM modelines in the headers help for auto indentation?

src/item.c
lwindolf: @rich-coe What do you think?

rich-coe: i'll take a look.

src/ui/item_list_view.c
rich-coe: @lwindolf I use vim as my ide.  I'm not sure that I see the indent issue you've identified in this snippet. 
I suspect it might have something to do with expandtab being turned on.

rich-coe: @lwindolf it is expandtab, I've fixed the spacing issue.

src/item.c
rich-coe: @lwindolf  adding timestr was an optimization to allocate and format the date string only once during the item fetch from the db, instead of every time the item gets rendered into the UI.

I've modifed it to only format the date string in item_list_view_update_item_internal().

PR #1002, opened 2021-06-04
Update pt_BR.po

fulvio-alves:


lwindolf:
Thank you for your translation update!

PR #1005, opened 2021-06-14
Fixed the `aria2` download selection not using the proper command

zorlaski:
The correct aria2 download command is `aria2c`.


lwindolf:
Thank you for the fix!

PR #1006, opened 2021-06-16
Add DHTML stripping tests.

lwindolf:
Also fix xml.c compiler warnings.

### DIFF
--- a/src/xml.c
+++ b/src/xml.c
@@ -243,7 +243,6 @@ typedef struct regex {
 
 static GSList *dhtml_strippers = NULL;
 static GSList *unsupported_tag_strippers = NULL;
-static GSList *expand_tag_regex = NULL;
 
 static void
 xhtml_regex_add (GSList **regex, const gchar *pattern, const gchar *replace)
@@ -271,7 +270,7 @@ xhtml_regex (const gchar *html, GSList *strippers)
 		GError *err = NULL;
 		regexPtr expr = (regexPtr) iter->data;
 		gchar *tmp = result;
-		result = g_regex_replace (expr->expr, tmp, -1, 0, expr->replace, 0, &err);
+		result = g_regex_replace_literal (expr->expr, tmp, -1, 0, expr->replace, 0, &err);
 		if (err) {
 			g_warning ("xhtml_strip: %s\n", err->message);
 			g_error_free (err);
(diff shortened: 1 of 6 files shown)


lwindolf:
@rich-coe Can you please have a look at this PR? I would like to re-add the script/iframe removal and have added tests to proof that it is really happening this time.

The test cases also uncovered some weaknesses in the regexes.

src/xml.c
```diff
@@ -148,7 +148,7 @@ xhtml_extract_doc (xmlNodePtr xml, gint xhtmlMode, const gchar *defaultBase)
 				}
 				xmlFreeDoc (oldDoc);
 			}
-			xmlFree (escapedhtml);
+			g_free (escapedhtml);
```
sunwire, line 151: Why variable escapedhtml is currently freed using the g_free() instead of xmlFree()?
It is allocated by [xmlNodeListGetString()](http://xmlsoft.org/html/libxml-tree.html#xmlNodeListGetString) and it should be deallocated with xmlFree()

src/xml.c
```diff
@@ -202,7 +202,7 @@ xhtml_extract_from_string (const gchar *html, const gchar *defaultBase)
 	if (body != NULL)
 		result = xhtml_extract (body, 1, defaultBase);
 	else
-		result = xmlCharStrdup ("<div></div>");
+		result = g_strdup ("<div></div>");
```
sunwire, line 205: If var body is NULL var result is allocated by g_strdup() otherwise by xmlCharStrdup() but they are always deallocated by xmlFree() see:
https://github.com/lwindolf/liferea/blob/71beee94a212387917a47e21f338b20e35e7c3ce/src/fl_sources/json_api_mapper.c#L154
https://github.com/lwindolf/liferea/blob/71beee94a212387917a47e21f338b20e35e7c3ce/src/fl_sources/ttrss_source_feed.c#L87

src/xml.c
lwindolf: There are 2 code paths one being xmlNodeListGetString() and one being g_strstrip(). This causes a type conflict, which is why the first code path casts to gchar. Therefore declaration and free'ing is of type gchar.

As for the background: you can configure libxml2 which allocator to use. And in Liferea it is configured to use Glib which is why gchar === xmlChar.

The reason why I changed the code above from xmlFree to g_free is to get rid of compiler warnings. Goal is to have a warning free compilation.

sunwire: > As for the background: you can configure libxml2 which allocator to use. And in Liferea it is configured to use Glib which is why gchar === xmlChar.

Ok, but if change an allocator to another one you have to remember to change g_free to a new function. If you forget this can lead to problems. 

> The reason why I changed the code above from xmlFree to g_free is to get rid of compiler warnings. Goal is to have a warning free compilation.

Is a type casting not enough? if I use xmlFree instead of g_free and xmlCharStrdup istead of g_strdup I don't see any warnings related to type conflicts.

PR #1007, opened 2021-06-16
Update codeql-analysis.yml

lwindolf:
Testing wether codeql does something useful when forcing C++


lwindolf:
Doesn't detect anything, but it is faster than the default config loading several minutes of python (probably due to our python based plugins).

PR #1018, opened 2021-07-29
Add pango_find_base_dir replacement

sunwire:
Because pango_find_base_dir replacement is deprecated I wrote a replacement for this function.
The code is based on the original function and is a combination of two functions pango_find_base_dir and pango_unichar_direction.
New function is named common_find_base_dir
fribidi library is now required to build and use.


lwindolf:
Thank you for fixing this deprecation!

PR #1019, opened 2021-07-30
Correction of an erroneous message

### LINKED ISSUE #1013: Certificate problems with some URLs
Some URLs like
https://www.php.net/releases/feed.php
Produce just "HTTP-Error 0: ".
I assume there is some problem/warning with the certificate the server uses.

sunwire:
It should resolve  #1013


sunwire:
@lwindolf I just noticed  that I accidentally added the po/pl.mo binary file. Please remove it. Sorry about that.

lwindolf:
Done. Don't worry. I didn't spot it either ;-)

PR #1021, opened 2021-08-19
Fix typo in "There are two option"

marciof:
Should be plural.


lwindolf:
Thanks for the typo fix!

PR #1022, opened 2021-08-30
Update fdm_pt.po

fulvio-alves:


lwindolf:
Thank you for the update!

PR #1023, opened 2021-09-03
Execute feed commands asynchronously

ittner:
Currently, running a command to generate feeds is a blocking operation
that freezes the interface and makes the entire program irresponsive.
This commit makes these commands run asynchronously in background
allowing interaction with the program while they are running.

An arbitrary 30s timeout is imposed to child processes to prevent slow
or blocked commands from running forever.

### DIFF
--- a/src/update.c
+++ b/src/update.c
@@ -282,6 +282,9 @@ update_job_new (gpointer owner,
 	job->flags = flags;
 	job->state = REQUEST_STATE_INITIALIZED;
 
+	job->cmd.fd = -1;
+	job->cmd.pid = 0;
+
 	return job;
 }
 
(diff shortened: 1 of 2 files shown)


lwindolf:
A lot of users will appreciate this, there were quite some reports on GUI freezes attributed to lacking legacy implementation.

I promise to have a look soon! Cheers.

lwindolf:
Works great! Thanks again for making this asynchronous!

ao2:
I just wanted to say thank you @ittner and @lwindolf 
I use [tweeper](https://ao2.it/143) a lot and and with this change my life rea..lly improved a lot :running_man:

lwindolf:
Thank you for the feedback!

PR #1027, opened 2021-09-27
Add option to always show news bins in the reduced feed list

ittner:
Add a new option "Always show in Reduced Feed List" to the news bin
properties so an user can make it appear there even when there are no
unread items. This enables some new and very practical workflows for
users wanting to use news bins to sort items or prepare reading lists
directly from the reduced view.

A new option always has a complexity cost for the user, but this one
seems worth it: the alternative of always showing news bins in the
reduced view, as happens with the search folders, can make the list
less useful for people who do not want all of them to be always there.

The setting is persisted in feedlist.opml, requires no changes to the
database schema and is forward and backwards compatible except, of
course, for the value being lost when loaded in a previews version.
Editing the option requires changes to the dialogs used to create news
bins and do not allow reusing the "rename node" dialog as a properties
dialog for this node type anymore. To prevent code duplication, the new
dialog is used both as a creating and a properties dialog.

### DIFF
--- a/glade/new_newsbin.ui
+++ b/glade/new_newsbin.ui
@@ -9,7 +9,6 @@
     <property name="modal">True</property>
     <property name="window_position">center-on-parent</property>
     <property name="type_hint">dialog</property>
-    <signal name="delete-event" handler="gtk_widget_hide_on_delete" swapped="no"/>
     <child>
       <placeholder/>
     </child>
@@ -32,7 +31,7 @@
                 <property name="can_default">True</property>
                 <property name="receives_default">False</property>
                 <property name="use_stock">True</property>
-                <signal name="clicked" handler="gtk_widget_hide" object="new_newsbin" swapped="yes"/>
+                <signal name="clicked" handler="gtk_widget_destroy" object="new_newsbin" swapped="yes"/>
               </object>
               <packing>
                 <property name="expand">False</property>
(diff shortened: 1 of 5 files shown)


ittner:
Folks,

I just updated this branch fixing a bug that only became obvious after using the patch for a few days (option was not set correctly when creating a new news bin).

For reference, the original branch is preserved in ittner/liferea:newsbin-always-show-in-reduced-list-before-fix ; the differences between the versions are the following:


```diff
ittner@olwe:liferea-contrib$ git diff newsbin-always-show-in-reduced-list-before-fix  newsbin-always-show-in-reduced-list
diff --git a/src/newsbin.c b/src/newsbin.c
index 6b395e08..2ec14828 100644
--- a/src/newsbin.c
+++ b/src/newsbin.c
@@ -106,6 +106,8 @@ on_newsbin_common_btn_clicked (GtkButton *button, gpointer user_data)
 
        if (newsbin == NULL) {
                newsbin = node_new (newsbin_get_node_type ());
+               node_set_data (newsbin, (gpointer)feed_new ());
+               newsbin_list = g_slist_append(newsbin_list, newsbin);
                newly_created = TRUE;
        }
 
@@ -114,8 +116,6 @@ on_newsbin_common_btn_clicked (GtkButton *button, gpointer user_data)
                ((feedPtr)newsbin->data)->alwaysShowInReduced = gtk_toggle_button_get_active (GTK_TOGGLE_BUTTON (showinreduced));
        }
        if (newly_created) {
-               node_set_data (newsbin, (gpointer)feed_new ());
-               newsbin_list = g_slist_append(newsbin_list, newsbin);
                feedlist_node_added (newsbin);
        }
 
```

lwindolf:
Sounds like a nice feature. Promise to review it in the next days!

lwindolf:
Works fine! Merged.

PR #1037, opened 2021-11-11
use g_time_zone_new_identifier instead of deprecated g_time_zone_new

sunwire:
Use g_time_zone_new_identifier when is available.


lwindolf:
Thank you for fixing this deprecation!

PR #1038, opened 2021-11-12
Simplified Chinese translation fixes

seflerZ:


lwindolf:
Thank you for providing this update of the Chinese translation!

seflerZ:
You're welcome. There're still lots to do. I'll continue to improve that.

PR #1040, opened 2021-11-13
Documentation - support for WCAG and HTML5 semantics

LorenzoAncora:
In addition to a professional look'n feel, those commits add basic support for accessibility devices and add the hooks required for more advanced CSS techniques. Some obsolete tags are replaced with semantic alternatives, the titles follow a meaningful hierarchy, semantic sections are specified and the behaviour of the hyperlinks is now uniform in all standard-compliant web browsers. Some enhancements to prevent future encoding errors are also included, because you never know.

### DIFF
--- a/doc/html/about_de.html
+++ b/doc/html/about_de.html
@@ -1,4 +1,4 @@
-<html>
+<html lang="de">
 <head>
 	<title>Einleitung</title>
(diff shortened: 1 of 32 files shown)


lwindolf:
Thank you for the great accessibility work and the Italian translation!

PR #1043, opened 2021-11-18
Full Italian translation of the Liferea Help's documentation

LorenzoAncora:
This PR translates all sections of the _Liferea Help_ into Italian.
Only the HTML documentation embedded within Liferea (which is accessible via _Help_ **&rarr;** _Summary_) is translated and the translated files are installed automatically. This is a manual translation made by a mother tongue and does not remove nor add information.


LorenzoAncora:
@lwindolf the translation is finally complete and should be ready for merge. ✅ 

Note: I suggest you should add locale detection to the Help (F1) command, so the new translation is used in place of the default English documentation when $`LANG` _starts with_ `it`. (this translation is compatible with all variations of Italian)
All translated files end with `_it.html` to ease their programmatic usage.

lwindolf:
Thanks for this massive documentation work! It is very appreciated!

BTW when using F1 the method liferea_shell_add_html_tab() ensures that the proper locale document is loaded. Doesn't it work for you? I just tested it with the German locale and it is fine.

LorenzoAncora:
> Thanks for this massive documentation work! It is very appreciated!

You're welcome! :-)

>> Note: I suggest you should add locale detection to the Help (F1) command, so the new translation is used in place of the default English documentation _when $LANG starts with it_. (this translation is compatible with _all variations_ of Italian)
>
> BTW when using F1 the method liferea_shell_add_html_tab() ensures that the proper locale document is loaded. Doesn't it work for you? I just tested it with the German locale and it is fine.

Yes it works _for me_, but it uses a locale function to hint language, thus processing more data than necessary in unreliable way. This [quick old-school example](https://gist.github.com/LorenzoAncora/448eb5d3740157a53fb4112917a1809b) gives an hint on why changing the algorithm might prevent future issues and a possible standard solution. It is only a conjecture but at least take it into consideration. ;-)

lwindolf:
Hmmm Liferea's common_get_localized_filename() does call g_get_language_names() as pointed out by the example you've linked. It should behave (i.e. iterating over all locale name possibilities) exactly as expected. Or do I miss something?

LorenzoAncora:
> Hmmm Liferea's common_get_localized_filename() does call g_get_language_names() as pointed out by the example you've linked. It should behave (i.e. iterating over all locale name possibilities) exactly as expected. Or do I miss something?

Not obvious but present: its semantics expect a well-formed set of ISO 639-1 but g_get_language_names returns an arbitrary (unpredictable in length, homogeneity and string format) list of ISO 639-1 + ISO 3166-1 alpha-2, probably thought for other use cases. Depending on system configuration, I think consequences could range from unnecessary disk access to mismatching or wrongly prioritized matching.
I've commented the example for you and linked one of the possible solutions at the bottom.

lwindolf:
@LorenzoAncora I finally understood (I think). So we should filter out useful/expected output of g_get_language_names().

LorenzoAncora:
Exactly, filter and sort the returned etherogeneous list to obtain a homogeneous set of two-letter ids or just use the standard locale functions directly as shown in 2nd example. Both methods are reliable.

LorenzoAncora:
@lwindolf I took the liberty to quickly code a patched version of `common_get_localized_filename` to solve the issue, see #1061. This isn't the best level of optimization (you'd need to implement set theory as described) but nonetheless should also improve performance a bit. **Merry Xmas**. 👋🏻 🎄 🎁

PR #1044, opened 2021-11-20
Support for user-agent string customization and anonymization

LorenzoAncora:
This PR adds the ability to fully customize the HTTP/S user-agent string used by Liferea.

- Power users can now impose an arbitrary User Agent string via the new `LIFEREA_UA` environment variable.
  Example:  ```$ LIFEREA_UA="My Arbitrary User Agent" ./liferea```
&hellip;replaces the user agent in its entirety with an arbitrary string;
- It is now also possible to define the new environment variable `LIFEREA_UA_ANONYMOUS` which will randomize the default user-agent.
 Example:  ```$ LIFEREA_UA_ANONYMOUS=1 ./liferea```
&hellip;randomizes the version number of Liferea and hides the OS identifier, respecting current standards;
- `LIFEREA_UA` has precedence over `LIFEREA_UA_ANONYMOUS`, so the latter can be safely defined at system level;
- The `--debug-net` parameter is now capable of printing the user agent which will be used by Liferea during each session.

There is no change to the normal user experience and no performance impact.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -3,6 +3,7 @@
  *
  * Copyright (C) 2007-2021 Lars Windolf <lars.windolf@gmx.de>
  * Copyright (C) 2009 Emilio Pozuelo Monfort <pochu27@gmail.com>
+ * Copyright (C) 2021 Lorenzo L. Ancora <admin@lorenzoancora.info>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


lwindolf:
Nice and simple addition. We have to think about some documentation too otherwise interested users won't find this feature.

LorenzoAncora:
> Nice and simple addition. We have to think about some documentation too otherwise interested users won't find this feature.

@lwindolf the end users capable of using environment variables are very likely to read the `man` pages. If you see fit, I can add a description of the new environment variables and of the currently undocumented `--debug-net` there, in the appropriate sections. Furthermore, adding a `EXAMPLES` section to hold a couple of examples of usage is also feasible.

PR #1046, opened 2021-11-22
Export source in node_source_export

mhoran:
When exporting subscriptions, ensure that the subscription is exported in addition to the node source. Otherwise, feed preferences won't be saved on change until the next sync, among other things.


lwindolf:
Good idea!

PR #1047, opened 2021-11-23
Various fixes to reader mode

### LINKED ISSUE #1045: "Refactor reader mode handling" breaks internal browser
I was syncing some patches I have to send a PR and noticed that the internal browser is broken for me on master. I bisected and found that 6d694e7bcb0b6b0308141422672289ce87950f04 breaks the internal browser for me. What I see is the following:

When I start liferea, I see "undefined" in the internal browser window. If I load an item that launches in the internal browser, the content renders properly but then falls back to the feed content. Note that all is fine if I disable reader mode. See the below screenshots:

Screenshot from 2021-11-21 19-27-02

Screenshot from 2021-11-21 19-27-19

Screenshot from 2021-11-21 19-27-21

mhoran:
* Fix Readability.js parsing for non-feed content
* Load WebView stylesheet via user stylesheet
* Reload the WebView when reader mode is toggled
* Default reader mode checkbox based on preferences

~Should also toggle the reader mode when the preference is changed in realtime but I wasn't able to figure that out quite yet.~

Fixes #1045

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@
 *.lo
 *.lai
 *.so
+*.swp
+*.swo
 Makefile
 Makefile.in
 Makecache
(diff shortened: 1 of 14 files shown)


lwindolf:
Thank you for this deep level fix! It is VERY appreciated.

PR #1048, opened 2021-11-23
Simplify webkit handling

lwindolf:
Rationale:

- In the past we supported multiple browser plugins
- Nowadays only Webkit is left
- Having the old plugin API makes passing stuff harder
  and causes a lot of code boilerplate

Idea of new class purpose:

- HtmlView -> adresses all HTML rendering stuff (XSLT, Javascript, running Readability.js)
- LifereaHtmlView -> rename to LifereaBrowser, complex widget (with history, URL bar, reader toggle ...)
- LifereaWebView allows for all direct Webkit manipulation (zoom, URL clicks, applying reader mode...)

TODO

- [x] Delete old plugin API
- [x] Rename LifereaHtmlView -> LifereaBrowser
- [ ] Maybe get rid of webkit.c by moving initialization to LifereaWebView
- [x] Make LifereaShell mostly use LifereaWebView (in general all places that do not want to change content, but the widget)


src/webkit/webkit.c
```diff
@@ -563,31 +523,36 @@ liferea_webkit_default_settings (WebKitSettings *settings)
...
-liferea_webkit_new (LifereaHtmlView *htmlview)
+GtkWidget *
+liferea_webkit_new (LifereaBrowser *htmlview)
 {
 	WebKitWebView 	*view;
 	WebKitSettings	*settings;
 
+	if (!liferea_webkit) {
+		liferea_webkit = g_object_new (LIFEREA_TYPE_WEBKIT, NULL);
+		liferea_webkit_init (liferea_webkit);
```
mhoran, line 539: I think this is superfluous; `g_object_new` calls init. I am getting a critical warning that the liferea scheme is being registered twice unless I remove this line.

lwindolf: You are right, fixed with 20340425

PR #1049, opened 2021-11-24
Update and extend the man page

LorenzoAncora:
This PR complements #1044 by documenting the new environment variables and a debug option previously undocumented.
In addition, it fixes existing grammar and increases the compatibility of the processed output.
The commits are structured so it will be easy to further extend the man page by simply repeating the existing patterns.

Note: an Italian translation is not included here. It will be included with #1043.


lwindolf:
Thank you for this rework!

PR #1050, opened 2021-12-05
Simplified Chinese translation update

seflerZ:


lwindolf:
Thank you for the zh_CN update!

PR #1053, opened 2021-12-07
Use liferea_shell_show_window when reactivating

mhoran:
Instead of calling gtk_window_deiconify and gtk_window_present directly, use the wrapper liferea_shell_show_window which also moves the window to the current desktop and restores the position if the window is hidden.

Note that we no longer restore position if iconified. I am not sure why this was ever the case but it seems unnecessary these days. Restoring position from iconification also triggers a bug on Cinnamon, whereas letting the window manager handle this seems to work fine.

Fixes an issue where position is not restored when Liferea is running in the background and is then launched from the launcher (instead of the tray icon.)

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -72,7 +72,7 @@ liferea_dbus_set_online (LifereaDBus *self, gboolean online, GError **err)
 static gboolean
 liferea_dbus_subscribe (LifereaDBus *self, const gchar *url, GError **err)
 {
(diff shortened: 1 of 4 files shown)


lwindolf:
To be honest this is almost impossible to test due to the multitude of window managers and X vs Wayland.

What you say makes sense and it seems to work for me on Ubuntu. So I think we should merge and wait for any users shouting :-)

PR #1057, opened 2021-12-11
Reset reader mode when writing directly to browser

mhoran:
We reset the reader mode when launching a URL, but should also reset when writing content directly to the browser (e.g. loading feed item content.)


lwindolf:
True!

PR #1058, opened 2021-12-18
fix it man installer

sunwire:
The commit f02014d  causes the DESTDIR variable is ignored and the Italian manual is always installed in /usr/share/man
This causes that it's not possible to prepare, e.g. a RPM package


lwindolf:
Thanks for fixing this!

PR #1061, opened 2021-12-23
Faster and safer generation of localized filenames

LorenzoAncora:
This PR:
- fixes a bug where `ISO 639-1` + `ISO 3166-1 alpha-2` + `encoding` identifiers (and variations) were accepted in place of pure `ISO 639-1` identifiers;
- adds the automatic conversion of upper case and mixed case language names into lower case language names, to allow for compatibility with case-sensitive filesystems when system locale names are misconfigured and/or partially configured;
- adds debugging to show which filenames are opened, selected or discarded;
- replaces code shorthands and semantically ambiguous names with documented and self-documented alternatives.

The consequence of this quick fix is a reduction in disk I/O and an overall stability improvement.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -1,9 +1,10 @@
 /**
  * @file common.c common routines for Liferea
- * 
+ *
  * Copyright (C) 2003-2013  Lars Windolf <lars.windolf@gmx.de>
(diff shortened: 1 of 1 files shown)


LorenzoAncora:
@lwindolf please feel free to make any adjustments before merging.

lwindolf:
Looks good! Thanks for the improvement!

PR #1069, opened 2022-01-03
Don't update HTML view when browsing internally

### LINKED ISSUE #289: Auto-loaded item link replaced with embedded content in internal browser during update
I have a few feeds configured to load items via the internal browser for sites that don't provide useful embedded content. When reading such an item during an update cycle, the browser reloads and displays the embedded content, replacing the auto-loaded item link. The issue persists until the update cycle is complete.

mhoran:
Fixes issues with the internal browser content being replaced with rendered feed items under various conditions.

Reproduction steps in https://github.com/lwindolf/liferea/issues/289#issuecomment-1004395375. To reproduce the mark as read issue, simply mark feed as read instead of updating.

Fixes #289.


lwindolf:
Works fine!

PR #1071, opened 2022-01-07
Various fixes and feature enhancements for TheOldReader sources

### LINKED ISSUE #548: Renamed subscriptions in TheOldReader not reflected in Liferea
If I rename a subscription in TheOldReader, that name will not be reflected in Liferea until I delete it (from Liferea). On the next sync, the updated name will be displayed.

mhoran:
* Syncs local node title with title from TheOldReader.
* Syncs node removal withTheOldReader (previously nodes would be deleted locally, but reappear on next refresh.)
* Sync local node folder change with TheOldReader. Introduces a new node source capability and implements on TheOldReader source.

I have tested these changes extensively with TheOldReader, but I do not have a Reedah account. Reedah support today doesn't have support for folders/categories, so those changes should be a no-op, and the delete change should also be safe.

Fixes #548


lwindolf:
Great work. Tested it and works like a charm!

PR #1073, opened 2022-01-19
Free previous stylesheet and only reload on change

mhoran:
There was a memory leak in the previous implementation any time render_get_css was called (which as it turns out can be often.)

Also, there's no need to reload the stylesheet if it hasn't changed. Unfortunately there's no way to get the stylesheet back from WebKit, so we keep a copy of it.

Also fix a method name that regressed with a previous refactor.

### DIFF
--- a/src/render.c
+++ b/src/render.c
@@ -322,7 +322,7 @@ render_is_dark_theme (void)
 	return darkTheme;
 }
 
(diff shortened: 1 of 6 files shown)


lwindolf:
Thanks again!

PR #1074, opened 2022-01-19
Pass settings to screen_changed signal handler

mhoran:
Fixes an assert that settings is not null.


lwindolf:
Thanks for the fix!

PR #1078, opened 2022-01-28
getfocus plugin use XDG_CONFIG_HOME path when is set

### LINKED ISSUE #1077: Get Focus plugin fails to save/read settings under Flatpak
Steps:

1. Run the flatpak `flatpak run net.sourceforge.liferea`
2. Enable Get Focus plugin
3. Change the opacity value, save and restart Liferea
4. See that the setting is not saved and no setting file is created.

I haven't tested this but I think the issue is here

https://github.com/lwindolf/liferea/blob/69cdf4c0ca820353131994732272c4395a438b45/plugins/getfocus.py#L36-L37

It tries to get the config directory relative to home directory but under the sandbox it is: `XDG_CONFIG_HOME=/home/user/.var/app/net.sourceforge.liferea/config`. (check with flatpak run --command='sh' net.sourceforge.liferea then printenv)

So it tries to save the setting to `/home/user/.config/liferea/plugin/...` but the sandbox has no access there, so it fails silently.

Instead it should try to save config relative to `XDG_CONFIG_HOME` as set in the environment.

sunwire:
Resolves #1077 
This PR allows you to save the plugin config file under Flatpak

### DIFF
--- a/plugins/getfocus.plugin
+++ b/plugins/getfocus.plugin
@@ -5,4 +5,4 @@ IAge=2
 Name=Get focus!
 Description=An automatic visual "focus mode" for Liferea, that makes the feedlist less prominent by change its opacity.
 Authors=Paweł Marciniak
-Copyright=Copyright © 2021 Paweł Marciniak
+Copyright=Copyright © 2021-2022 Paweł Marciniak
--- a/plugins/getfocus.py
+++ b/plugins/getfocus.py
@@ -1,7 +1,7 @@
 """
 GetFocus! Liferea plugin
 
-Copyright (C) 2021 Paweł Marciniak <sunwire+liferea@gmail.com
+Copyright (C) 2021-2022 Paweł Marciniak <sunwire+liferea@gmail.com
 
 This program is free software: you can redistribute it and/or modify
 it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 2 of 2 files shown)


bbhtt:
Thank you!  Tested the Flatpak and it works.

If someone wants to test, download flatpak bundle from https://github.com/bbhtt/net.sourceforge.liferea/blob/external-plugins/net.sourceforge.liferea.flatpak (built from [this](https://github.com/bbhtt/net.sourceforge.liferea/tree/external-plugins) branch) `flatpak install --user net.sourceforge.liferea.flatpak`

sunwire:
> Can you please do the Pane position plugin within this as well, if possible? Then only the plugin downloader will remain.

At least I can try but please open separate issues for pane plugin and plugin installer it makes easier to track and fix them

sunwire:
@bbhtt I simplified the get_path code if you can also test it, I would be grateful

bbhtt:
> At least I can try but please open separate issues for pane plugin and plugin installer it makes easier to track and fix them

Ok, I separated the issues.

> @bbhtt I simplified the get_path code if you can also test it, I would be grateful

Yep, still works.I updated the flatpak bundle as well.

Thanks.

sunwire:
> Yep, still works.I updated the flatpak bundle as well.

OK. Thanks

PR #1081, opened 2022-01-29
pane plugin use XDG_CONFIG_HOME path when is set

### LINKED ISSUE #1080: Pane position plugin fails to save/read settings under Flatpak
(Splitting from the main issue) 

Steps:

1. Run the flatpak `flatpak run net.sourceforge.liferea`
2. Enable Pane psotion workaround plugin
3. See that no setting file is created in `/home/user/.var/app/net.sourceforge.liferea/config/liferea/plugins`

The issue is probably here:

https://github.com/lwindolf/liferea/blob/69cdf4c0ca820353131994732272c4395a438b45/plugins/pane.py#L34-L35

It tries to get the config directory relative to home directory but under the sandbox it is: `XDG_CONFIG_HOME=/home/user/.var/app/net.sourceforge.liferea/config`. (check with ` flatpak run --command='sh' net.sourceforge.liferea` then `printenv`)

So it tries to save the setting to `/home/user/.config/liferea/plugin/...` but the sandbox has no access there, so it fails silently.

Instead it should try to save config relative to `XDG_CONFIG_HOME` as set in the environment.

sunwire:
Resolves #1080
@bbhtt Test will also be welcome.

### DIFF
--- a/plugins/pane.plugin
+++ b/plugins/pane.plugin
@@ -5,4 +5,4 @@ IAge=2
 Name=Pane position workaround
 Description=Set pane position, workaround.
 Authors=Paweł Marciniak
-Copyright=Copyright © 2021 Paweł Marciniak
+Copyright=Copyright © 2021-2022 Paweł Marciniak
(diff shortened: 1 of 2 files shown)


bbhtt:
Thanks, works. It can correctly create and write to the `pane.conf` file in `/home/now/.var/app/net.sourceforge.liferea/config/liferea/plugins/pane` now.

PR #1082, opened 2022-01-29
plugin-installer use XDG_DATA_HOME path when it is set

### LINKED ISSUE #1079: Plugin installer copies to wrong path under Flatpak
(Splitting from the main issue)

Steps:

1. The Flathub flatpak doesn't include `git` so this should be used instead: https://github.com/bbhtt/net.sourceforge.liferea/blob/external-plugins/net.sourceforge.liferea.flatpak `flatpak install --user net.sourceforge.liferea.flatpak` and `flatpak run net.sourceforge.liferea`
2. Install a plugin
3. See that in terminal that it is copying from `/tmp/...` to `~/.local/share/liferea/plugins` but the sandbox has no access to `~/.local/share/liferea/plugins` so plugin installation doesn't persist on restart.

The issue is probably here:

https://github.com/lwindolf/liferea/blob/69cdf4c0ca820353131994732272c4395a438b45/plugins/plugin-installer.py#L72 

It should use `XDG_DATA_HOME`, under the sandbox it is set to `XDG_DATA_HOME=$HOME/.var/app/net.sourceforge.liferea/data` and if not set it falls back to ` $HOME/.local/share`

sunwire:
Should resolve #1079
@bbhtt Please test it.

### DIFF
--- a/plugins/plugin-installer.py
+++ b/plugins/plugin-installer.py
@@ -19,6 +19,7 @@
 
 import urllib.request, json
 import os, sys, glob, shutil, subprocess
+from pathlib import Path
 import gi
 
 gi.require_version('Gtk', '3.0')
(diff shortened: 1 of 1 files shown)


bbhtt:
Thanks, works! I updated the [bundle](https://github.com/bbhtt/net.sourceforge.liferea/blob/external-plugins/net.sourceforge.liferea.flatpak) with all three patches now. With this installing the external plugins from Liferea becomes possible under the Flatpak sandbox.

lwindolf:
Solid solution!

PR #1102, opened 2022-03-17
[WIP] Load raw content in Readability.js instead of DOM loading

### LINKED ISSUE #1101: Refactor reader mode: load raw content, avoid DOM load
First loading the actual website and then applying reader mode
- takes longer
- runs 3rd party script before we control the page
- eats more resources
- often waits in a cookie wall

Instead of injecting Readability.js on a loaded DOM let's render a default view with takes an asynchronously network fetched HTML snippet and passes it to Readability.js. This will be

- much faster
- loads no 3rd party CSS, JS, front, ... resources
- sees no cookie walls

lwindolf:
- More secure, because no DOM evaluation happens
- Opens up the possibility to use DOMPurify.js for further stripping
- Avoids loading any JS, cookie walls, ... upon document.load
- But will break Reader mode for all pages that dynamically insert content

Goal here is to be resource efficient, fast and secure with the
trade-off of not loading all pages perfectly.

Closes #1101

### DIFF
--- a/js/htmlview.js
+++ b/js/htmlview.js
@@ -32,19 +32,39 @@ function loadContent(readerEnabled, content) {
 	}
 	if (true == readerEnabled) {
 		try {
-			console.log('[liferea] enabling reader mode...');
+			console.log('[liferea] reader mode is on');
+			var contentDiv = document.getElementById('content');
 			var documentClone = document.cloneNode(true);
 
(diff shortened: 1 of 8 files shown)


lwindolf:
Todos left:

[ ] Toggling reader mode in internal browser now broken
[ ] Solve issue of internal URI scheme causing no URL bar being visible

PR #1110, opened 2022-04-18
Update Dutch translation

Gert-dev:
Gave the Dutch translation some love; it seems it hasn't been updated since 2013 according to the PO file.

I limited myself to adding missing translations and updating fuzzy ones.


lwindolf:
Great contribution! Thanks a lot!

PR #1114, opened 2022-05-04
Add generic Google Reader API support.

lwindolf:
Add a generic Google Reader API feed source that can be used with FeedHQ, Miniflux, FreshRSS.

Later on investigate whether to switch TheOldReader, Inoreader and Reedah implementations to this implementation.

Decisions
- restore old code
- drop older OPML API subscription list fetching for Google Reader (as e.g. Miniflux only wants to speak JSON)
- make JSON output mandatory for several callbacks 
- do feed fetching via /stream/items/ids and /stream/items/contents (todo: check if possible for Reedah)

Todos 
- [x] restore old code
- [x] implement JSON quick update
- [x] debug output JSON on subscription update
- [x] debug/fix Miniflux error on edit-tag
- [x] fix using name field as node title from subscribing dialog
- [ ] fix crash when API token is not yet available (e.g. on login fail + background update)
- [x] refactor current "Google Reader API" source to (for the user) separate source "Miniflux", "FeedHQ" and "FreshRSS", to do so make facade implementations each with own source ids, so we can later on separately deprecate them if needed
- [x] add /stream/items/ids support
- [x] add /stream/items/contents support
- [x] fix feedlist_save_cb() crash
- [x] fix add feed crash
- [x] fix remove feed crash
- [x] clarify wether uppercase 'T' is a valid POST data field


lwindolf:
Miniflux bug prevents marking items read https://github.com/miniflux/v2/pull/1437/files

lwindolf:
- Item fetching works when testing with TheOldReader.
- Does not work with Miniflux (see https://github.com/miniflux/v2/issues/1498)

sjehuda:
I suggest to rename Google Reader into G-Reader, as I don't think they deserve to have their name mentioned.

sjehuda:
This post be deleted

PR #1117, opened 2022-05-11
Fix strange left padding of headline row

### LINKED ISSUE #1115: Strange left padding in headline colomn
There is a padding space in the first column. Don't know its function and how to hide it.
image

seflerZ:
fix #1115


lwindolf:
Thank you for contributing this!

PR #1128, opened 2022-07-27
Various improvements to appdata

bbhtt:
This improves the display of Liferea on Software Stores and Software Centers.

Some of it is done automatically by `appstream-util upgrade` https://github.com/flathub/flathub/wiki/AppData-Guidelines#upgrading like the metadata license, the rest is done manually.

This can be validated with `appstream-util validate-strict net.sourceforge.liferea.appdata.xml`

Changes:

- Adds a copyright header https://github.com/flathub/flathub/wiki/AppData-Guidelines#header
- Uses the new id tag system https://github.com/flathub/flathub/wiki/AppData-Guidelines#id, provides and launchable
- Added a developer_name tag: 
developer_name
- Added update_contact, please let me know if you don't want the email displayed there. The mailing list was removed since it is no longer available.
- Updated the feature list section with https://lzone.de/liferea/ :
 
app_description

- Added all screenshots from https://lzone.de/liferea/screenshots.htm and added the captions. The width/height property was removed since it is not necessary. The first image is marked as default (All of them doesn't need to be the default image). 
- Added bugtracker and donation links:
 
links

- Added translations as per https://github.com/flathub/flathub/wiki/AppData-Guidelines#translations

- Added changelog from https://github.com/lwindolf/liferea/releases/tag/v1.13.9: 

changelog

Unfortunately formatting like bold is not supported so it looks a bit verbose.

This is the only section that needs updating whenever a new release is made, before the release, so that it is in the release tarball. The release version/date is mandatory and the changelog is optional (but nice to have). It'd be nice if @lwindolf can do it in the release workflow.

### DIFF
--- a/net.sourceforge.liferea.appdata.xml.in
+++ b/net.sourceforge.liferea.appdata.xml.in
@@ -1,42 +1,80 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2022 Lars Windolf  -->
 <component type="desktop-application">
-  <id>net.sourceforge.liferea.desktop</id>
-  <metadata_license>CC0</metadata_license>
+  <id>net.sourceforge.liferea</id>
+  <provides>
+    <id>net.sourceforge.liferea.desktop</id>
+  </provides>
+  <launchable type="desktop-id">net.sourceforge.liferea.desktop</launchable>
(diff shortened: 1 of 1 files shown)


lwindolf:
Thank you for this update!

PR #1132, opened 2022-08-07
Improved timeout error handling in feed commands

ittner:
I was facing some timeout issues with feed-generating commands that may be affecting other users too so follows three commits with small  improvements. They are separate changes, but also closely related and mutually dependent, so they all go in a single PR.

First commit allows overriding it via environment variable `LIFEREA_FEED_CMD_TIMEOUT`, in the same way Liferea already uses `LIFEREA_UA` for the user-agent string; second increases the default timeout to 60 s (it seems I was too optimistic with the 30 s timeout when I first added it), and the third commit changes the error reporting so it becomes possible to tell timeouts and other errors apart.

I hope they will be useful!

### DIFF
--- a/man/it/liferea.1
+++ b/man/it/liferea.1
@@ -94,6 +94,13 @@ usato come URI del proxy. Entrambe vengono usate da molti strumenti CLI, quindi
 assicurati di esportarle in una sotto-shell dedicata.
 .RE
 .TP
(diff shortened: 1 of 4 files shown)


ittner:
Quick update to the PR, adding another commit with descriptive strings for HTTP server errors (status codes 5xx).

lwindolf:
PR looks good.

PR #1133, opened 2022-08-16
Add search rule "podcast" that only checks for audio enclosures

ittner:
Add a new search rule for podcasts that check if the feed has at least
one enclosure with MIME type "audio/*". This removes the false positives
the old rule caused by assuming that any enclosure was a podcast
(including videos and images that some feeds may have).

Changing the rule would have changed the behavior of the saved searches
that users already have, so the new rule is added with a different id and
the existing one is kept with the existing id, but with a different title
("Enclosure") that better describes its actual behavior. This preserves
compatibility and allows users to use the old rule if they need it.

### DIFF
--- a/src/rule.c
+++ b/src/rule.c
@@ -25,6 +25,7 @@
 #include "common.h"
 #include "debug.h"
 #include "metadata.h"
+#include "enclosure.h"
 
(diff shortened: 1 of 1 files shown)


lwindolf:
Good idea!

PR #1139, opened 2022-09-18
 Save enclosures when exporting items to a file 

ittner:
Also export the enclosure elements when exporting items to a feed file. This is mostly useful for podcasts.

Please notice that this PR already includes the commit to fix the invalid memory access to function enclosure_from_string() that I sent a few minutes ago in PR #1138. To make things easier, you can merge this one and close both issues.


lwindolf:
Thank you for the improvement and the bugfix!

PR #1143, opened 2022-09-26
Update default feed sources

ittner:
Use HTTPS for a few more URLs in the default feeds and update the default sources for Brazilian Portuguese.


lwindolf:
Great work!

PR #1144, opened 2022-09-28
Fix issue 901 - Panes position are not properly restored

### LINKED ISSUE #901: Layout not remembered

sunwire:
Fixes #901 - Panes position are not properly restored
Removes pane position plugin.


src/ui/liferea_shell.c
```diff
@@ -1161,7 +1159,9 @@ liferea_shell_restore_state (const gchar *overrideWindowState)
 		gtk_container_child_set (GTK_CONTAINER (liferea_shell_lookup ("wideViewPane")), liferea_shell_lookup ("wideViewItems"),
 			"resize", TRUE, NULL);
 	}
-	
+
```
lwindolf, line 1162: Cannot merge this. Window position is only safe when set from within the idle callback.

lwindolf (changes requested):

src/ui/liferea_shell.c
sunwire: Ok, I understand.

PR #1146, opened 2022-10-03
Fix color scheme support across desktop environments

ittner:
A sequence of commits that fix a few issues with the color scheme support for different desktop environments: a memory leak after reading org.gnome.desktop.interface.color-scheme, an always-false comparison for org.freedesktop.appearance.color-scheme and a crash if the DE has no support for dark themes at all (which also happens with any Gnome < 42). 

In commit 078d32ddd006a89b5fe9701f66ee3dc225fb84ae I assumed that the ideas was for Liferea to get the dark theme by default if user sets no preference (from both the typo being || → && and from the preference stated in commit 72e9567bbc05f84f79822b164f5f6a40a3d99eb3 ... please point if I was wrong here).

I hope I haven't missed anything as I haven't tested in many DEs.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -81,20 +81,24 @@ conf_get_dark_theme (void)
 	if (fdo_settings) {
 		gint scheme;
 
-		conf_get_int_value_from_schema (fdo_settings, "color-scheme", &scheme);
-		debug1 (DEBUG_CONF, "FDO reports color-schema code '%d'", scheme);
-		if (1 == scheme)
-			dark = FALSE;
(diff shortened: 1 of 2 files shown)


ittner:
I recreated the branch to fix a bug when unref'ing the returned schema. Also updated the branch for this PR, so everything will work as intended.  For reference, the original branch (with the bug) is preserved in [here](https://github.com/ittner/liferea/tree/fix-color-scheme-support-across-des-orig).

lwindolf:
Thanks for this great debugging work!

PR #1147, opened 2022-10-03
Use UTF-8 encoding for feedlist.opml

ittner:
Use UTF-8 instead of ASCII for feedlist.opml, so minimizing the amount of XML entities there for users who have lots of non-ASCII characters in their feeds, which is specially useful if they also edit the feed list manually with a standard text editor.

Liferea will load this file file with any encoding supported by libxml2, so this change is both backward and forward compatible.


lwindolf:
Nice improvement!

PR #1149, opened 2022-10-09
Improvements for podcast search rule

ittner:
Two small improvements for the podcast search rule that I only noticed when writing the new author search rule for PR #1148 : make code shorter (by using `metadata_list_get_values` instead of callbacks) and add the documentation to the help files.


lwindolf:
Thanks for also providing documentation!

PR #1150, opened 2022-10-12
Update pt-BR translation

ittner:
Noticed the RC1, so this is a good time to check this translation! I should have noticed the missing strings before, but I haven't edited a po-file for a while and don't even use my systems in Portuguese... anyway, it's updated now.

Also pinging @fulvio-alves , who made most recent translations.


lwindolf:
Thanks for updating!

PR #1170, opened 2022-11-03
Fix notification icon in libnotify plugin

tatokis:
Commit 0dea39b79802 changed the filename of the liferea icon but the plugin was not updated to reflect that.


lwindolf:
Thanks for fixing this!

PR #1172, opened 2022-11-14
Fix fuzzy strings for pt-BR translation

ittner:
Replace two instance of fuzzy-matched strings by their correct versions. I somehow missed these in 81b87963d43f437e01de2671432fa3c83c7b60d7.


lwindolf:
Thanks!

PR #1176, opened 2022-11-20
Load custom user stylesheet for all URLs

### LINKED ISSUE #1163: Custom liferea.css no longer working
Testing out the latest RC I found that my custom `liferea.css` no longer works. I had to revert both 4d911024934949bc9f012eecbb8e64717d69758b and 6ddbf9372a47de1541a51bb9060f302704dc39ae to get it working again.

Also, editing my `liferea.css` used to automatically update the stylesheet currently applied in Liferea. I now have to quit and restart in order to get it to apply. IIRC there used to be some signal that would trigger `render_get_css` when the `liferea.css` was modified.

I use a custom `liferea.css` to hide various nonsense in the internal browser. Not sure how to best go about fixing this as the above changes look related to the dark theme which I'm unfamiliar with.

mhoran:
Only load the default stylesheet, used for theming, when loading Liferea-generated HTML.

Fixes #1163.

### DIFF
--- a/src/render.c
+++ b/src/render.c
@@ -302,8 +302,7 @@ render_get_theme_color (const gchar *name)
 {
 	GSList	*iter;
 
-	if (!themeColors)
-		return NULL;
+	g_return_val_if_fail (themeColors != NULL, NULL);
 
 	iter = themeColors;
 	while (iter) {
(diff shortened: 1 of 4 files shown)


lwindolf:
Works great and the code complexity went down. Perfect!

PR #1177, opened 2022-11-24
Update CN translation

Davidy22:
Updating this while waiting for 1.14, I eagerly await the arrival of 1.14 in my distro's repos. Tried to stick to words for things already used in the existing translations for things like feed, news bin etc

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -6,285 +6,258 @@
 #
 msgid ""
 msgstr ""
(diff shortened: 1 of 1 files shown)


Davidy22:
Closing for now because I worked off a translation from a tarball like it recommended but it seems like it's quite different from the translations in git master, will do some copypasting and updating and reopen

PR #1178, opened 2022-11-24
Update CN translation

Davidy22:
Updating this while waiting for 1.14, I eagerly await the arrival of 1.14 in my distro's repos. Tried to stick to words for things already used in the existing translations for things like feed, news bin etc.

Second attempt at this, messed up with the last one and started from what I think was an old .po file because i went to a tarball instead of the git repo first.


lwindolf:
Thank your for this great work. Updating such an old translation is a lot of work!

Davidy22:
Just eager to get a version with the timestamp fix in my distro repo, doing what I can to help it along.

PR #1183, opened 2022-12-16
Update Spanish translation

cfuga:
Update Spanish translation for Liferea


lwindolf:
Thank you for this translation update!

Issue #1068, opened 2022-01-03
[Refactor] Drop all proxy handling in Liferea


lwindolf:
To simplify our code base drop all proxy preferences and environment logic as this is properly covered by libsoup nowadays (see GProxyResolver docs).

Change documentation to hint on desktop environment settings or (for non-standard users) to use environment variables to allow for full proxy flexibility.

RedHoodHacker:
Hey is it okay if I take a crack at this? I know Python and Bash and I know that isn't relevant right now but I'll
do what this says and then ask you to take a look at it? I am a Noob at opensource projects and want to get into them to work on coding and such...

lwindolf:
Turned out to be unrealistic.

PR #1184, opened 2022-12-16
Update liferea_webkit_set_proxy to use non-deprecated functions

mhoran:
I was holding off on this but given #1068 is closed, this PR addresses deprecation warnings for proxy handling. I've tested this the best I can though I don't regularly use a proxy -- so if someone who does regularly use a proxy can give it a further test that'd be great.


lwindolf:
Given that the change is an almost simple replace I consider the risk low.

Thank you for fixing those deprecations!

PR #1207, opened 2023-02-22
Use monotonic clock in libnotify plugin

tatokis:
It is a good idea to use a monotonic clock as a way to check how much time has passed, as it will not be affected by the system clock changing (DST for example).

However, I am not sure what the minimum targeted Python 3 version is.

From the [docs](https://docs.python.org/3/library/time.html#time.monotonic) for `time.monotonic()`:
```
New in version 3.3.

Changed in version 3.5: The function is now always available and always system-wide.

Changed in version 3.10: On macOS, the function is now system-wide.
```

If the minimum Python version is < 3.5, then a check needs to be added to use `time.monotonic()` if it's available.

### DIFF
--- a/plugins/libnotify.py
+++ b/plugins/libnotify.py
@@ -56,7 +56,7 @@ class LibnotifyPlugin(GObject.Object, Liferea.ShellActivatable):
         self.shell.props.feed_list.disconnect(self._handler_id)
 
     def on_node_updated(self, widget, node_title):
-        new_timestamp = time.time()
+        new_timestamp = time.monotonic()
 
         # Only make a new notification every 10 seconds
         if new_timestamp > self.timestamp + 10:


lwindolf:
This improvement makes sense. I guess the Python 3.5 requirement is not a problem as we depend on WebkitGtk4+ which on no system I know of is available with older Python version.

So I'd say let's merge this and maybe just add something like this to the plugin

    import sys
    MIN_PYTHON = (3, 5)
    assert sys.version_info >= MIN_PYTHON, f"requires Python {'.'.join([str(n) for n in MIN_PYTHON])} or newer"

lwindolf:
@tatokis What do you think about the suggestion above?

tatokis:
@lwindolf Sounds good to me.

Sorry, I was thinking of scrapping this entirely and instead of timing it, just clearing the message when the notification is dismissed (as there seems to be a closed event, but I haven't actually tested it, and I have the suspicion some implementations might not implement it correctly).

PR #1208, opened 2023-02-22
Add MessagingMenu plugin

tatokis:
A long time ago I used to use Liferea on Ubuntu 16.04. The version packaged in the repositories carried a C patch that added Liferea to Unity's MessagingMenu.
It was very convenient and useful, and I was extremely frustrated when I had to upgrade to 20.04 and my workflow was broken (as the patch was dropped, probably because it was difficult to maintain).

With that in mind, I've been trying to make an easy to maintain Python plugin that restores this functionality.

The current version is in a decent enough state (I've been using it daily for a quite while now.), however I would like some feedback as it feels like I might be going down the wrong path.

As such, a few questions came up:

- Is there a better way to access the feeds in the order they are shown in the TreeView?
  - It needs to still possible to make Liferea switch to that feed like the plugin currently does in `_source_activated()`
- If not, is there a better way to filter out the non feed entries (such as Unread, Important, and folders)?

TODO:
- [ ] Handle Gtk TreeView/TreeModel item reordering
  - Any kind of reordering/adding a new feed will currently break things and end up with duplicate entries
- [ ] Add config window
  - [ ] Add configurable option to show new entries after start (and not existing unread items)
  - [ ] Add configurable option to order MessagingMenu entries based on when they were last updated
- [ ] Show default RSS icon when a feed does not have a favicon available

I am probably forgetting a few things right now, so I'll update this when I come across them again.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -60,7 +60,9 @@ plugin_DATA = \
 	plugins/plugin-installer.py \
 	plugins/plugin-installer.plugin \
 	plugins/trayicon.py \
-	plugins/trayicon.plugin
+	plugins/trayicon.plugin \
+	plugins/messagingmenu.py \
+	plugins/messagingmenu.plugin
 
 gsettings_SCHEMAS = net.sf.liferea.gschema.xml
 @INTLTOOL_XML_NOMERGE_RULE@
--- a/plugins/messagingmenu.plugin
+++ b/plugins/messagingmenu.plugin
@@ -1,0 +1,8 @@
+[Plugin]
+Module=messagingmenu
+Loader=python3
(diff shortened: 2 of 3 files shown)


lwindolf:
@tatokis Regarding your questions:

1. Is there a better way to access the feeds in the order they are shown in the TreeView? -----> No there is not. This is because the nodePtr type is not yet an GObject and so there is no serialization to access it. I plan to change it to GObject in the future allowing access to it.
2. If not, is there a better way to filter out the non feed entries (such as Unread, Important, and folders)? ---> right now not, when nodePtr becomes accessible you could simple query the feed type.

PR #1225, opened 2023-03-30
Draft: Upgrade to libsoup3

lwindolf:
Experiment on how to convert to libsoup3.

When used with plugins that use `gi.require` 4.0 Webkit this crashes Liferea with

      (WebKitWebProcess:61150): libsoup-ERROR **: 22:14:21.803: libsoup3 symbols detected. Using libsoup2 and libsoup3 in the same process is not supported.

This is a disaster. Upgrading users will just see Liferea "crash" with a weired error message that does not tell them what to do.

Todos before this can be used

- [ ] Prevent Webkit related plugins to crash Liferea
- [x] Wait for Github Action runner upgrade to Ubuntu 23.04, right now we cannot CI-build due to libsoup3.2 being only in Ubuntu 22.10 while GH offers only 22.04


rezso:
You need to change webkit2gtk-web-extension-4.0 to webkit2gtk-web-extension-4.1 in the configure.ac
(https://github.com/lwindolf/liferea/blob/master/configure.ac#L62)
With this change, I can compile the source with this patch :)

lwindolf:
@rezso Thank you for this hint. This solves the non-plugin related issue. 

What remains is that every Webkit2GTK 3 using plugin still prevents Liferea from starting.

Running a `grep "gi.require_version('WebKit2', '4.0'" $HOME/.local/share}/plugins` to auto-disable all plugins that do this is not a nice solution.

rezso:
I cannot find any gi.require_version('WebKit2', '4.0') in the plugins.

lwindolf:
@rezso It is mostly the plugins from mozbugbox for example: https://github.com/mozbugbox/liferea-plugin-studio/blob/1e97019e55daff0053b00319a6cc2a7ebda4fdae/webkit-setting/webkitsetting.py#L42

lwindolf:
My current idea is to provide a method that the plugin can ask for the Webkit version Liferea uses and to only require this version.

rezso:
The one release of mozbugbox/liferea-plugin-studio is 8 years old, so I think, this is not your problem :)
And a hint: because libsoup3 support exists since webkit 2.33.2 only, you can remove checks and support of older webkit versions.

lwindolf:
@rezso The webkit version checks are needed to match the plugins to the proper webkit version as multiple versions can be installed (e.g. Webkit 4.0 and 4.1 on Ubuntu 22.10). Selecting the wrong one per-default will break features or cause the problem I mentioned.

Also as maintainer I consider the mozbugbox/liferea-plugin-studio features important and widely used. As those plugins are directly downloaded by the plugin installer there are no releases used.

uduecoder:
I'm not sure I really get the plugins argument, but I'd like to share my point of view. For me, the problem is not at runtime but at compile time already. On my LinuxFromScratch, I'd like to link liferea against webkitgtk41 rather than 40, because it is not trivial to get the old one to build with current tool chain. Switching to webkitgtk41 will result in the libsoup2vs3 issue - at compile time already. Therefore I'm patching liferea since last September like this:
sed -e 's@-4.0@-4.1@' -e 's@-2.4 >= 2.42@-3.0@' -i configure.ac
patch -Np1 -i ../liferea-1.13.9_libsoup3.patch
I did not speak up earlier, because I'm not fully confident that patch is correct, I'm not too familiar with libsoup. It does the job for me, however updating multiple times in a row may result in not getting the desired result for all sources, as if I failed to use proper libsoup3 parameters for caching/sessions/whatever. But maybe it inspires someone to do it correctly - it is not all that much:
[liferea-1.13.9_libsoup3.patch](https://github.com/lwindolf/liferea/files/11142153/liferea-1.13.9_libsoup3.patch)

yselkowitz:
Please note that [Fedora 40 will drop support for webkit2gtk-4.0](https://fedoraproject.org/wiki/Changes/Remove_webkit2gtk-4.0_API_Version), so liferea needs to be updated to libsoup-3.0 and webkit2gtk-4.1 soon in order to remain in Fedora.

genodeftest:
It seems like the github builder is running Ubuntu 22.04 LTS according to [the build logs](https://github.com/lwindolf/liferea/actions/runs/4583711234/jobs/8094728751?pr=1225#step:1:4), which has `libsoup3` version 3.0.x installed but your changes to `configure.ac` require `libsoup3 >= 3.2.0`.

Would it be fine to downgrade libsoup requirement to 3.0 for the build to work in Ubuntu 22.04 LTS? Or would it be an option to use a newer version of Ubuntu? Or maybe enable some backports repo which might have libsoup3 3.2 or newer?

genodeftest:
> Wait for Github Action runner upgrade to Ubuntu 23.04, right now we cannot CI-build due to libsoup3.2 being only in Ubuntu 22.10 while GH offers only 22.04

According to the [Github blog entries related to Github Actions](https://github.blog/changelog/label/actions/), it seems like Github only updated to LTS versions, i.e., it won't update to a version of Ubuntu fulfilling your needs before Ubuntu 24.04 LTS, i.e. in about a year. Fedora 40 is expected to be released a bit earlier, so it would be too late for that.

Is there another option?
* Is `libsoup3 3.0` an option?
* Would it be fine for you to switch to other images, not the ones supplied by Github, but from Docker's container registry:  https://hub.docker.com/_/ubuntu ?

lwindolf:
@genodeftest I'll try to do this. I did not use GH actions so far with non-managed images. Seems like it is possible to do so https://github.blog/changelog/2020-09-24-github-actions-private-registry-support-for-job-and-service-containers/

lwindolf:
Given the distros "running away" version wise I think it is better to upgrade now, even if the plugin crash issue remains. I guess we'll have to deal with it via support.

bbhtt:
> Given the distros "running away" version wise I think it is better to upgrade now, even if the plugin crash issue remains. I guess we'll have to deal with it via support.

I doubt it ever worked, since some of the functions have long been removed from webkitgtk https://github.com/mozbugbox/liferea-plugin-studio/blob/1e97019e55daff0053b00319a6cc2a7ebda4fdae/webkit-setting/webkitsetting.py#L385C24-L385C51

```
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.11/site-packages/gi/module.py", line 126, in __getattr__
    raise AttributeError("%r object has no attribute %r" % (
AttributeError: 'gi.repository.WebKit2' object has no attribute 'get_default_session'
```

with webkit2gtk 2.40.2

uduecoder:
I appreciate the effort. Master builds and runs on a webkitgtk41/libsoup3 system without patching now. However, it shares the problem I hinted at for my initial hack from last September:

At least after multiple attempts to "Update All" about half of my sources (that work fine the first time) show an error sign and clicking on it the following warning box:
There was a problem when fetching this subscription!
✅ 1. Authentication
⛔ 2. Download
⬜ 3. Feed Discovery
⬜ 4. Parsing
Details:

--

I don't see a clear pattern, may or may not work with/out atom or rss, http or https, but some sources stay fine,
e.g. http://www.anandtech.com/rss/
while other sources persistently run into this error,
e.g. https://xkcd.com/atom.xml .

Neither knowing liferea nor libsoup internals well enough, I have trouble to identifier what the problem with my initial libsoup2->3 migration would be in that regard.

Can someone more enlightened please fix repeated libsoup3 download of e.g. https://xkcd.com/atom.xml !?

PR #1228, opened 2023-04-01
Update appdata for 1.14.4

bbhtt:
1.14.4 tag was created from `liferea-1_14` branch which is not synced with the latest changes from master. So the release is missing the previous appdata updates. Not sure if this was intentional. It's not an issue though :)

### DIFF
--- a/net.sourceforge.liferea.appdata.xml.in
+++ b/net.sourceforge.liferea.appdata.xml.in
@@ -105,6 +105,15 @@
     <content_attribute id="money-gambling">none</content_attribute>
   </content_rating>
   <releases>
+    <release date="2023-03-21" version="1.14.2">
+      <description>
+        <p>This is a stability fix for 1.14.1</p>
+        <ul>Changes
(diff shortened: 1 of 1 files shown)


lwindolf:
@bbhtt This was not intentional. The question is more in what branch should the appdata changes go?

bbhtt:
> @bbhtt This was not intentional. The question is more in what branch should the appdata changes go?

master - if that's the branch you are going to make releases from. Ideally we want the previous changelogs in the appdata preserved so that frontends (GNOME Software/KDE) can show them. Right now on the 1.14.4 tag, the appdata is missing the changes for 1.14.2 and 1.14.3. It's not a major issue as the Flathub web frontend cannot display historical version info.

lwindolf:
Future 1.14 releases will happen from the `liferea-1_14` branch. The master branch is for ongoing development. Liferea does not do GitFlow, but trunk-based development.

So I think the appdata merge requests need to go into `liferea-1_14 branch` too.

Issue #1203, opened 2023-02-05
Feature proposal: a new "flat view" mode and filter by feed title


ittner:
I am proposing a new view mode for Liferea that shows all feeds in a single level, similar to the reduced view mode but also showing the feeds with no unread items, and a text box to filter feeds by their titles. Idea is show the filter in the new reduced view too, as it seems useful there too, specially for folks with lots of subscriptions.

The "Reduced view" checkbox in the view menu is replaced by a "Normal/Reduced/Flat radio buttons" (feedback is welcome for these view names!) :

![New menu options replacing the 'reduced view' checkbox](https://user-images.githubusercontent.com/110642/216849429-27c00c55-6d37-47d0-9bbe-99b272f6ebec.png)

And the new filter box :

![Flat view mode with no filter](https://user-images.githubusercontent.com/110642/216849439-d3eddab6-864a-48d6-838b-4b386085a175.png)

![Flat view mode filtering only BBC feeds](https://user-images.githubusercontent.com/110642/216849446-a2f4bf13-8684-4bab-8557-1d24915c83be.png)

I chose to call this box as  "Filter" so it is not confused with the usual search feature, but I would also like some feedback on this.

I still wondering if it is a good idea to also filter by URL: it seems very practical (type "youtube" and get all channels, "reddit" and all subreddits, etc.), but it may also be confusing for some users, as the base for the results is not visible.

I have a working proof-of-concept in https://github.com/ittner/liferea/tree/flatview-and-title-filter-poc -- this code stills need polishing and documentation, but I am already using it for a while and it also seems good enough for discussion and tests among advanced users (i.e. people who install Liferea from source); If this feature is welcome in Liferea, I plan to continue the development and prepare a proper PR for merge.

PR #1249, opened 2023-06-13
Add new flat view mode and title filter

ittner:
I originally sent this as a discussion issue ( #1203 ), but it was more than four months ago and the feature already proved itself valuable for me and I didn't find any bug. So, it's time for me to bite the bullet with a PR =)

This PR introduces a new view mode for Liferea that shows all feeds in a single level, similar to the reduced view mode but also showing the feeds with no unread items, and a text box to filter feeds by their titles. The filter is also shown in the new reduced view, as it seems useful there too, specially for folks with lots of subscriptions (I subscribe to ~520 feeds ATM and already miss this kind of filter).

The "Reduced view" checkbox in the view menu is replaced by a "Normal/Reduced/Flat view" radio buttons (but feedback is welcome for these names!) :

New view mode menu

And the new filter box :

Filter box with no text, showing all entries

Filter box listing only BBC subscriptions

I chose to call this box as "Filter" so it is not confused with the usual search feature, but I would also like some feedback on this.

### DIFF
--- a/glade/liferea_menu.ui
+++ b/glade/liferea_menu.ui
@@ -153,8 +153,19 @@
        </section>
       <section>
         <item>
-          <attribute name="action">app.reduced-feed-list</attribute>
-          <attribute name="label" translatable="yes">_Reduced Feed List</attribute>
+          <attribute name="action">app.feedlist-view-mode</attribute>
+          <attribute name="label" translatable="yes">N_ormal feed list</attribute>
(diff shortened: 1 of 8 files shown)


ittner:
Hi! Does this feature fit your plans for LIferea? I plan to send more patches again now that I have a bit of extra time again, but I'm not sure if it will be worth to update this branch to the latest master for a possible merge.

PR #1251, opened 2023-06-13
Fix incorrect item_id when downloading AMP URLs

ittner:
Code passed an incorrect element when enriching an entry from a AMP URL, resulting in a pointer to the item becoming the id. This commit fixes the issue.


lwindolf:
Good catch! Thanks for fixing!

PR #1253, opened 2023-06-18
Preserve search rule argument when changing rule types

ittner:
Do not erase the rule argument when user changes a rule type in the editor used for the search folders and advanced search. This is a purely usability change, no effect on persistent data.

Current behavior:

https://github.com/lwindolf/liferea/assets/110642/e267cf56-b499-4c5d-96a2-eca553b4e8a1

New behavior:

https://github.com/lwindolf/liferea/assets/110642/da0bd06d-7222-4eab-9ccf-2cbdbcbe1ec4


lwindolf:
Thanks for this usability fix!

PR #1255, opened 2023-06-22
Defer removing read items also for middle-clicking to mark as read.

dperelman:
Related to #788/#997: defer removing items from view for reasons other than them being selected. The specific problem I had that this solves is that this makes it so middle-click to mark an item as read will not immediately remove the item from a folder/search view of unread items.


dperelman:
It looks this behavior hasn't changed, although the code it targets has changed slightly, so the PR code will have to be modified. I will update and reopen the PR soon.

lwindolf:
That would be awesome!

lwindolf:
@dperelman Yes, this is unintended and will be fixed. The current main is the work in progress GTK4 port which is quite unfinished.

PR #1262, opened 2023-06-28
Plugins installer: Remove duplicate punctuation (translatable string)

genodeftest:


genodeftest:
Note: This change obviously modifies strings to be translated.

lwindolf:
@genodeftest I'll additionally patch all translations not to have the dot. So we can add this to 1.15

lwindolf:
Will be included in 1.14.6 and 1.15.0

PR #1278, opened 2023-08-02
Update pt_BR.po

fulvio-alves:
Please add this Portuguese (Brazil) translation update.


lwindolf:
Thank you for this update!

PR #1284, opened 2023-08-09
updated czech translation

Amereyeu:


lwindolf:
Thank you for this update!

Issue #1285, opened 2023-08-12
Subscription Update with no available update signals an error


rich-coe:
when update fetches a subscription url and the return status is 304 with 0 bytes received, the subscription is in the error state.

rich-coe:
Found during investigating #1276

rich-coe:
created pull request

When a subscription is requested to update, and it hasn't updated since the last time in the current liferea session,
and the status is 304, liferea was setting the subscription into the error state.

lwindolf:
PR merged. Will be included in 1.15.1

PR #1286, opened 2023-08-12
Check for 304 status before 0 length data #1285

rich-coe:
When a subscription is requested to update, and it hasn't updated since the last time in the current liferea session,
and the status is 304, liferea was setting the subscription into the error state.

Check for 304 status before checking for no data.


lwindolf:
Thank you for debugging and fixing this!

PR #1292, opened 2023-08-17
Fix spurious word break

paulgevers:
commit 64179f85b seems to have introduced two spurious spaces, remove them again.


lwindolf:
Thanks for fixing!

Issue #1227, opened 2023-03-31
[1.14.4] corrupted double-linked list


paulgevers:
I think the 1.14 branch is cursed :(.

I have added a `dogtail` based regression test to my Debian package [1]. I took 1.14.4, build it in the usual Debian way, installed the packages and ran the test on my system. It hang the system around the time it tried to open the `About` window. When I started `liferea` from the command line, I got this when I tried to do it manually. I think timing wise the `killed` part is probably my manually sending `sigkill`.
```
paul@mulciber ~ $ liferea

(liferea:302405): Gtk-WARNING **: 21:17:58.096: Theme parsing error: <data>:2:28: The style property GtkButton:default-border is deprecated and shouldn't be used anymore. It will be removed in a future version

(liferea:302405): Gtk-WARNING **: 21:17:58.096: Theme parsing error: <data>:3:36: The style property GtkButton:default-outside-border is deprecated and shouldn't be used anymore. It will be removed in a future version

(liferea:302405): Gtk-WARNING **: 21:17:58.096: Theme parsing error: <data>:4:25: The style property GtkButton:inner-border is deprecated and shouldn't be used anymore. It will be removed in a future version

(liferea:302405): Gtk-WARNING **: 21:17:58.096: Theme parsing error: <data>:5:30: The style property GtkWidget:focus-line-width is deprecated and shouldn't be used anymore. It will be removed in a future version

(liferea:302405): Gtk-WARNING **: 21:17:58.096: Theme parsing error: <data>:6:27: The style property GtkWidget:focus-padding is deprecated and shouldn't be used anymore. It will be removed in a future version
corrupted double-linked list
Killed
```

I tried to run with `--debug-all` but that's a bit much (can send out-of-band). It ended with:
```
HTML   : Error running javascript: undefined:1:8: ReferenceError: Can't find variable: setBase
GUI    : Enabling reader mode for 'liferea://'
HTML   : loadContent() reader mode fail -> reloading without reader
CONF   : Setting last-window-state to 0
CONF   : Setting last-window-state to 0
CONF   : Setting last-window-state to 0
CONF   : Setting last-vpane-pos to 261
CONF   : Setting last-hpane-pos to 199
CONF   : Setting last-wpane-pos to 428
CONF   : Forced feed list save
CACHE  : adding vfolder: title=Important
CACHE  : adding vfolder: title=Unread
CACHE  : adding feed: source=https://xkcd.com/atom.xml interval=-1 cacheLimit=
[...]

PR #1295, opened 2023-08-23
Fix two license claims

paulgevers:
While trying to package 1.15.1 for Debian I noticed 

1) The license claim of plugins/add-bookmark-site.py had similar problems as #950, I propose to bring it in line with the other plugins
2) The header of src/tests/ui-test.py didn't reflect what josch and I intended, as can be seen in the comments of #1227, the license is supposed to be GPL2+, not LGPL2+.


lwindolf:
@paulgevers Sorry about the wrong info in src/test/ui-test.py!!! Thanks for creating the PR!

PR #1309, opened 2023-10-11
Add Ayatana AppIndicator backend to trayicon plugin

nE0sIghT:
Since GNOME 3.26 Shell do not show tray icons and `Gtk.StatusIcon` is deprecated for some time.   

There were several GNOME extensions that implements system tray and actual extension is [AppIndicator and KStatusNotifierItem Support](https://extensions.gnome.org/extension/615/appindicator-support/). This extension supports both Freedesktop's System Tray Protocol and StatusNotifierItem specification.  

The problem with System Tray Protocol is that it doesn't works with Wayland obviously. And for some reasons Liferea's `Gtk.StatusIcon` looks weird for me [1].  

With this change StatusNotifierItem icon [2] is added to trayicon plugin through [Ayatana Application Indicator](https://github.com/AyatanaIndicators/libayatana-appindicator) and made preferred tray icon implementation if libayatana-appindicator is available.

[1] 
Screenshot from 2023-10-11 21-57-46

[2] 
Screenshot from 2023-10-11 23-09-17


lwindolf:
Will try this out soon. Sounds like a good idea, although it is hard to get an idea how common Ayatana indicators are among the user base.

nE0sIghT:
I added minor change to support legacy Canonical Appindicator too (Ayatana appindicator is the fork).  
As per Repology [1][2] both the legacy appindicator and the ayatana-appindicator are widelly adopted by Linux distros.  

In my Gentoo box it is used by NetworkManager applet and Remmina. It looks like it's used by Transmission too.

[1] https://repology.org/project/libappindicator/versions
[2] https://repology.org/project/libayatana-appindicator/versions

plugins/trayicon.py
```diff
@@ -26,6 +26,18 @@
 from gi.repository import GObject, Gtk, Liferea
 from gi.repository import Gdk, GdkPixbuf
 
+gi.require_version('AyatanaAppIndicator3', '0.1')
```
lwindolf, line 29: The gi.require not being safe-guarded by the `try` below causes a 

     ValueError: Namespace AyatanaAppIndicator3 not available

for users without Ayatana GIR bindings. So it needs to be included in the `try` and both `except` need to be extended to

     except (ImportError, ValueError)

lwindolf (changes requested):
Other than the problem noted works great!

plugins/trayicon.py
nE0sIghT: I pushed the fix. Thanks!

PR #1312, opened 2023-10-20
trayicon: workarounds for AppIndicator issues

nE0sIghT:
It should be noted that for some DEs with StatusNotifier support (eg Mate [1]) the look of attention icon will change and can not be reverted. However it will be consistent between X11 and Wayland sessions.  I think this is ok and look and feel enhancement requests should be filled for such DEs if needed.

We can split off AppIndicator support into separate plugin instead of this workarounds however code needs to be duplicate between 2 plugins.

[1] The look of attention icon in Mate
image


lwindolf:
For simplicity I think one plugin is better. I consider the modernisation of the code very important and especially the improved visibility of the counter is worth any incompatibilities.

Issue #1258, opened 2023-06-27
Can we add oneline revision signature in About?


stefanos82:
When we build a stable release it should indeed show

![image](https://github.com/lwindolf/liferea/assets/1777466/a135bbb8-8c28-4952-b3aa-1cb483e97e44)

but when we pull the latest changes from the repo and build the code ourselves, we should get something like

![image](https://github.com/lwindolf/liferea/assets/1777466/a08c44e7-3629-4efc-8905-41580adba770)

Is it plausible to add this feature?

I'm asking because if I hit a bug, I won't be able to know which commit revision is responsible for the cause of buggy behavior so I can report it.

lwindolf:
Would need some extensive automake magic.

stefanos82:
Hmm, well with cmake you can do it as follows:

```cmake
elseif(GIT_FOUND)
    execute_process(
        COMMAND ${GIT_EXECUTABLE} rev-list --max-count=1 HEAD
        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
        OUTPUT_VARIABLE TR_VCS_REVISION
        OUTPUT_STRIP_TRAILING_WHITESPACE)
endif()
```

Code snippet directly taken from https://github.com/transmission/transmission/blob/main/CMakeLists.txt#L154-L160

Logically with makefile you can run something like `$(shell git rev-list --max-count=1 HEAD)` and will give you the same output as above.

lwindolf:
Sorry, I decided not to add this.

PR #1319, opened 2023-10-24
Revision signature in About

bbhtt:
Not very familiar with autoconf/automake but the idea should work. I've tried to keep it very simple.

Note. 

- The version string now needs to be only updated in `.version`. The tarball will be named `liferea-v$version` if from a tag or `liferea-$sha-$date` if from commit.
- I don't know it is preferred to keep the "v" in version or not.
- I haven't checked inside a tarball but it should just work.
- This needs backported to 1.14 branch if merged with a different `.version` file.

Fixes https://github.com/lwindolf/liferea/issues/1258

<details>
<summary> Images </summary>

From the current HEAD git commit that does not belong to a tag: 

5986e03ddb

From a checked out tag https://github.com/bbhtt/liferea/releases/tag/v1.110.819: 

cfd58e2868


bbhtt:
The issue is intlool-update doesn't support m4_esyscmd_s but does m4_esyscmd. I wanted to use the latter to avoid newlines. I guess I'll add stripping instead.

version.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env sh
+
+set -eu
```
bbhtt, line 3: I wonder if this makes sense, I think if git is not available in a build system it'll make the script fail. Maybe I should remove this altogether or reverse the conditional to check for the non git case first.

lwindolf:
Sorry for taking so long to review.

I'm very unsure whether the solution would work well, as I cannot predict the build enviroment of package maintainers. They more or less rely on configure.ac to statically provide a version. And we cannot predict wether they'll build from a git checkout or using a fresh source directory (make dist).

bbhtt:
In case it is not from git, it reads the first line of the `.version` file using `head`. 

Instead of the static version in configure.ac it will be now in `.version`.

I don't think that will be an issue?

PR #1322, opened 2023-10-27
Remove Mate workaround from trayicon plugin

nE0sIghT:
Issue was fixed in Mate and under fix in Fedora.

See-Also: https://github.com/mate-desktop/mate-panel/issues/1412
See-Also: https://bugzilla.redhat.com/show_bug.cgi?id=2245224


sunwire:
Tested, works as expected (Fedora 38 mate-panel-1.26.3-3)

PR #1329, opened 2023-12-31
src/node_type.h: add `#include <libxml/tree.h>` for `xmlNodePtr` type

trofi:
Without the change build fails on `libxml2-2.12.3` as:

    ../../src/node_type.h:62:64: error: unknown type name 'xmlNodePtr'; did you mean 'nodePtr'?
       62 |         void            (*export)               (nodePtr node, xmlNodePtr cur, gboolean trusted);
          |                                                                ^~~~~~~~~~
          |                                                                nodePtr

### DIFF
--- a/src/node_type.h
+++ b/src/node_type.h
@@ -23,6 +23,7 @@
 
 #include "node.h"
 
+#include <libxml/tree.h>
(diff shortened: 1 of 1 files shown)


lwindolf:
Thanks!

PR #1339, opened 2024-02-24
Update for sq translation

ujdhesa:


lwindolf (approved):
Thanks for the update!

PR #1352, opened 2024-04-02
Allow building with libsoup 3.0.7

ittner:
This version has all the feature currently required and will allow building on Ubuntu 22.04 without any other changes.


lwindolf:
Thanks for testing this and lowering the dependency!

PR #1372, opened 2024-07-03
Fix feed icon not parsing if the URL doesn't have a file extension

Stomern:
When parsing a feed that contains an icon URL without a file extension, [this line](https://github.com/lwindolf/liferea/blob/db561b20c78f4a41c0567e08922fb2f1cb92e10c/src/subscription_icon.c#L165C1-L165C50) gets called to search for favicons which don't exist if the response is an image, leaving the icon blank like in the screenshot below. This change moves the original logic for deciding which function to call for checking the icon URL to a new callback function and adds an extra check to download the image if the content type in the response matches a valid image mime type.

Example feed: https://invidious.fdn.fr/feed/channel/UC5UAwBUum7CPN5buc-_N1Fw (Invidious feed for The Linux Experiment Youtube channel)
Icon URL: https://yt3.googleusercontent.com/ytc/AIdro_lAKf-vZLoTI-gZUoP5Y3gbdGd07E4eDHUhTee6aOzDCnU=s900-c-k-c0x00ffffff-no-rj

image

### DIFF
--- a/src/subscription_icon.c
+++ b/src/subscription_icon.c
@@ -128,6 +128,23 @@ subscription_icon_download_html_cb (const struct updateResult * const result, gp
 }
 
 static GRegex *image_extension_match = NULL;
+static GRegex *image_mime_type_match = NULL;
+
+static void
+subscription_icon_handle_response (const struct updateResult * const result, gpointer user_data, updateFlags flags)
+{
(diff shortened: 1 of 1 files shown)


lwindolf:
Thank you for this fix!

PR #1398, opened 2024-12-11
Bug1276 use libcurl instead of libsoup

rich-coe:
Attempt to resolve lost connections from libsoup using libcurl.
Missing proxy support.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -38,6 +38,7 @@ pkg_modules="	gtk+-3.0 >= 3.24.0
 		sqlite3 >= 3.7.0
 		gmodule-2.0 >= 2.0.0
 		gthread-2.0
+                libcurl >= 8.9.0
 		libsoup-3.0 >= 3.0.7
 		webkit2gtk-4.1
 		json-glib-1.0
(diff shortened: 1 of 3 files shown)


lwindolf:
@rich-coe Sorry for not reacting!

I'm mostly concerned about the testing, so I'd say this should not go into 1.15/1.16. What do you think? Or would you say libsoup3 is too unstable for a 1.16.0 release? In my daily usage I see rare bugs, but there are almost invisible overall.

So if at all I'd tend to switch the networking together with the Gtk4 port, which needs a lot of testing anyway.

lwindolf:
Don't worry about any merge conflicts. They are not hard to resolve :-)

rich-coe:
@lwindolf 
I don't have any issues waiting for the GTK-4 integration.
The only two issues I know of are (1) determining if the proxy is configured and fetching the configuration info for the proxy, and (2) not marking the feed 'in error' on the UI when some feeds return a non-success code for no update since last referesh.

rich-coe:
for (1) proxy config, the code currently fetches the proxy info.  I'll have to setup the proxy config and test that it works.
for (2) I'll debug the feed that this happens on.

lwindolf:
Sorry, I didn't expect the MR to get closed when performing a master/main switch :-(

rich-coe:
@lwindolf  It's ok.  I'll probably have to create a new PR against branch gtk4.

PR #1411, opened 2025-02-10
Update Spanish (es) translation

cfuga:
Signed-off-by: Cristian Othón Martínez Vera <cfuga@cfuga.mx>


lwindolf:
Thank you for the update!!!

PR #1416, opened 2025-02-25
Polish translation update

sunwire:


lwindolf:
Thank you for the update!

PR #1429, opened 2025-04-19
Migrate to girepository 2 0

lwindolf:


lwindolf:
Currently fails with https://gitlab.gnome.org/GNOME/gobject-introspection/-/issues/548

configure.ac
```diff
@@ -74,11 +74,10 @@ AC_MSG_RESULT(user agent OS = $uname)
 GOBJECT_INTROSPECTION_CHECK([0.9.3])
```
loqs, line 74: Is GOBJECT_INTROSPECTION_CHECK not for gobject-introspection-1.0 which links to girepository-1.0 which causes `Warning: cannot register existing type 'GIRepository'` because girepository-1.0 is already loaded in the process preventing the loading of girepository-2.0?

lwindolf: I was also wondering about this. As girepository-2.0 only supports meson, the build instructions for autotools are unclear. 

Build-wise gobject-introspection is still necessary as I understand as not all tools moved to glib2 (containing girespository2).

So in the end the new code AFAIK has 2 deps: gobject-introspection and girepository-2.0

loqs: I think so from looking at the meson changes to pygobject for [a18dd9b WIP: Port to libgirepository-2.0 ](https://gitlab.gnome.org/GNOME/pygobject/-/commit/a18dd9b0d35f41e81213a6b3e4ff095691f824c5) which introduced the incompatibility. @heftig is this correct?

heftig: Pretty sure autotools is unsupported.

heftig: The tools to generate gobject introspection metadata are (and remain) in `gobject-introspection`, which provides `GOBJECT_INTROSPECTION_CHECK`. I think this should continue working...

heftig: The INTROSPECTION_LIBS AND INTROSPECTION_CFLAGS aren't used by https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/Makefile.introspection?ref_type=heads

loqs: @lwindolf if you override introspection libs with `make INTROSPECTION_LIBS='-lgirepository-2.0 -lgobject-2.0 -lglib-2.0'` after patching one remaining use of girepository-1.0 does that leave only girepository-2.0 loaded?
```
--- a/src/liferea_application.c
+++ b/src/liferea_application.c
@@ -44,7 +44,7 @@
 #include "xml.h"
 #include "ui/liferea_shell.h"
 
-#include <girepository.h>
+#include <girepository/girepository.h>
 
 struct _LifereaApplication {
 	GtkApplication	parent;
@@ -308,7 +308,7 @@ liferea_application_init (LifereaApplication *self)
 
 	g_application_add_main_option_entries (G_APPLICATION (self), entries);
 	g_application_add_option_group (G_APPLICATION (self), debug);
-	g_application_add_option_group (G_APPLICATION (self), g_irepository_get_option_group ());
+	g_application_add_option_group (G_APPLICATION (self), gi_repository_get_option_group ());
 
 
 	g_signal_connect (G_OBJECT (self), "activate", G_CALLBACK (on_app_activate), NULL);
```

lwindolf: @loqs Thanks for the suggestion. I applied the fix with [dfd1dc1](https://github.com/lwindolf/liferea/pull/1429/commits/dfd1dc100618bd0b98548c9d0d6a7ac98742a5fe)

When I compile like this I end up with a binary only linked to girepository-2.0 

    $ ldd src/liferea |grep girepo
            libgirepository-2.0.so.0 => /lib/x86_64-linux-gnu/libgirepository-2.0.so.0 (0x000077a82aece000)
    $

So everything should be fine, right?

loqs: Yes and as per @heftig I think you should be able to drop the use of INTROSPECTION_CFLAGS and INTROSPECTION_LIBS from src/Makefile.am src/tests/Makefile.am and hopefully all the libraries will then be able to coexist.

lwindolf: @loqs Yep, removing INTROSPECTION_CFLAGS and LIBS works, I need to figure out how to switch it for both compile options...

loqs: > I need to figure out how to switch it for both compile options...

I believe both options are unused irrespective of the setting of `USE_GI_REPOSITORY_VERSION` and `INTROSPECTION_COMPILER` and `INTROSPECTION_SCANNER` do not depend on the value of `USE_GI_REPOSITORY_VERSION`.

lwindolf:
FYI I'm merging this for 1.16, but will stay with default to girepository-1.0. So maintainers can change it depending on the pygobject.

Issue #1448, opened 2025-08-28
rss tests fail with libxml2 2.14.5


loqs:
## Version
liferea 1.16.2
libxml2 2.14.5

## Problematic Behavior
One of the rss tests fails as libxml2 has stopped implying `<p>` start tags. 

## Steps to Reproduce
```
$ git clone https://github.com/lwindolf/liferea
$ cd liferea
$ git checkout v1.16.2
$ autoreconf -fiv
$ ./configure
$ make
$ make test
....
# Start of rss tests
**
ERROR:parse_rss.c:85:tc_parse_feed: assertion failed (tc[i] == json): ("{\"id\":0,\"title\":\"i1\",\"description\":\"<div xmlns=\\\"http://www.w3.org/1999/xhtml\\\"><p>D</p></div>\",\"source\":\"http://localhost/item1.html\",\"readStatus\":false,\"updateStatus\":false,\"flagStatus\":false,\"time\":1678397817,\"validTime\":false,\"validGuid\":false,\"hasEnclosure\":false,\"sourceId\":null,\"nodeId\":null,\"parentNodeId\":null,\"metadata\":[],\"enclosures\":[]}" == "{\"id\":0,\"title\":\"i1\",\"description\":\"<div xmlns=\\\"http://www.w3.org/1999/xhtml\\\">D</div>\",\"source\":\"http://localhost/item1.html\",\"readStatus\":false,\"updateStatus\":false,\"flagStatus\":false,\"time\":1678397817,\"validTime\":false,\"validGuid\":false,\"hasEnclosure\":false,\"sourceId\":null,\"nodeId\":null,\"parentNodeId\":null,\"metadata\":[],\"enclosures\":[]}")
[...]

PR #1452, opened 2025-08-31
Add libxml2 2.14 support to parse RSS test

loqs:
- libxml2 removed implied `<p>` tags https://gitlab.gnome.org/GNOME/libxml2/-/commit/8cf6129bbd836e666e7eda8c9e61c00387ae388b.
- Only add `<p>` tags for libxml2 versions before 2.14 (< 21400).
- Related https://github.com/lwindolf/liferea/issues/1448.

### DIFF
--- a/src/tests/parse_rss.c
+++ b/src/tests/parse_rss.c
@@ -20,6 +20,7 @@
 
 #include <glib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


lwindolf:
Thank you for adding the version guard!

PR #1454, opened 2025-09-05
Fixes missing icon in headerbar plugin for mark all as read 

bitlord:
Uses alternative symbolic icon instead of one removed from default Gnome theme[1]

[1] https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/merge_requests/78


bitlord:
Hi, 

I'm not sure is this right or not, for me icon shows some default "icon" in the headerbar, this patch applied locally to 1.16.x works for me and displays checkmark icon as it should. The icon seems to be the same as original (symbol).

This is all tested only on default Gnome desktop (48) and default icon theme.

lwindolf:
Looks like a resonable icon!

PR #1468, opened 2025-10-31
Distribute js/gopher-renderer.js

yaneti:


lwindolf:
Arg... totally forgot to add it. And "make distcheck" does not catch it as it is used on runtime only.

Thanks for adding!

PR #1513, opened 2026-05-07
po: Add Georgian translation

EkaterinePapava:


lwindolf:
Thank you for the new Georgian translation!

Issue #1534, opened 2026-06-12
New liferea Icon not displayed in Window or Dock in 2.0-RC2 & 3


AndyM48:
<img width="1087" height="74" alt="image" src="https://github.com/user-attachments/assets/7c3d3ef0-8898-42e1-bfb6-eadaa2ce01cc" />

- The icon is not showing on the left of the title bar on my system or in the Dock

_Originally posted by @AndyM48 in https://github.com/lwindolf/liferea/discussions/1521#discussioncomment-17114754_

lwindolf:
I cannot reproduce this so far.

I assume you used "meson install" to install the compiled source. Can you please run 

    find /usr/local/share/icons/hicolor |grep liferea

and

    sudo /usr/bin/gtk4-update-icon-cache -q -t -f /usr/local/share/icons/hicolor

and provide the output?

AndyM48:
`find /usr/local/share/icons/hicolor |grep liferea`
I suspect that is the problem. I installed to /usr, so the icons are in /usr/share/icons/hicolor. Perhaps you have hard coded the icon folder?

```
$ find /usr/local/share/icons/hicolor |grep liferea
find: ‘/usr/local/share/icons/hicolor’: No such file or directory
```

```
$ sudo /usr/bin/gtk4-update-icon-cache -q -t -f /usr/local/share/icons/hicolor
gtk4-update-icon-cache: Failed to open file /usr/local/share/icons/hicolor/.icon-theme.cache : No such file or directory
```

lwindolf:
Well not hard-coded, but determined at compile time via config.h (PACKAGE_DATA_DIR). But I consider this a standard way to determine app specific paths.

AndyM48:
Well clearly it is not working. I installed using your meson instructions including `--prefix /usr` and the icons were correctly installed to /usr/share/icons/hicolor (actually they were already there from previous installs).

It seems that in spite of this, liferea expects the icons to be at  `/usr/local/share/icons/hicolor/`, otherwise, why else are the icons not showing up on my install?

If you like I could test this by copying the icon files to /usr/local and seeing if they show up?

AndyM48:
Hmm. I copied the icon files to /usr/local and updated the icon cache and the icons are still not shown on my /usr installation?
Probably too simple an answer.

lwindolf:
[...]

PR #1542, opened 2026-06-18
Implement trayicon with libstray

lwindolf:
Implementation limitations compared to old implementation:

- no support for new number count display next to icon
- no support for menu checkbox for window close behaviour
- single left click does not toggle window visibility anymore (double click does)

### DIFF
--- a/.github/workflows/cb.yml
+++ b/.github/workflows/cb.yml
@@ -26,7 +26,7 @@ jobs:
 
     - run: |
        apt-get update -qq
-       apt-get install -y -qq libxml2-dev libxslt1-dev libsqlite3-dev libwebkitgtk-6.0-dev libjson-glib-dev libgirepository-2.0-dev libpeas-2-dev gsettings-desktop-schemas-dev python3 libtool intltool valgrind libfribidi-dev gla11y appstream desktop-file-utils meson ninja-build gobject-introspection libadwaita-1-dev
+       apt-get install -y -qq libxml2-dev libxslt1-dev libsqlite3-dev libwebkitgtk-6.0-dev libjson-glib-dev libgirepository-2.0-dev libpeas-2-dev gsettings-desktop-schemas-dev python3 libtool intltool valgrind libfribidi-dev gla11y appstream desktop-file-utils meson ninja-build gobject-introspection libadwaita-1-dev libdbus-1-dev
        mkdir inst
        mkdir -pv inst/share/glib-2.0/schemas
        mkdir -pv inst/share/icons/hicolor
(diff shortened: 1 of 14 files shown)


AndyM48:
Does this only work in gnome?
What should I install on Openbox to enable it? I have libpeas-2 installed.

Edit: Error messages
```
(net.sourceforge.liferea:26293): libpeas-WARNING **: 12:14:39.655: Bad plugin file '/usr/lib/liferea/plugins/trayicon/trayicon.plugin': embedded plugins must be a resource

(net.sourceforge.liferea:26293): libpeas-WARNING **: 12:14:39.655: Error loading '/usr/lib/liferea/plugins/trayicon/trayicon.plugin'


```

lwindolf:
@AndyM48 The error message happens due to the old plugin file still in /usr/lib/... you'd need to remove it as it would try loading and conflict with the new C plugin.

As for non-GNOME I cannot say. I'd guess any Status-Notifier-Item providing library would suffice.

This is exactly what I was talking about: having to support what end users have to install on their Linux distros. This I cannot provide.

AndyM48:
OK, I understand the problems, please don't stress, I appreciate your hard work.

The trayicon messages have gone after deleting the plugin.

AndyM48:
Trayicon works on Openbox with Stalonetray. 
To implement, install snixembed  and run 
`$ snixembed --fork`
then run liferea.

Single-click on the status icon shows/hides window. Right-click shows menu, menu items both work.

Currently I have no icon, but this is due to my installing to /usr instead of /usr/local. (see #1534)

lwindolf:
> Does this only work in gnome? What should I install on Openbox to enable it? I have libpeas-2 installed.
> 
> Edit: Error messages
> 
> ```
> (net.sourceforge.liferea:26293): libpeas-WARNING **: 12:14:39.655: Bad plugin file '/usr/lib/liferea/plugins/trayicon/trayicon.plugin': embedded plugins must be a resource
> 
> (net.sourceforge.liferea:26293): libpeas-WARNING **: 12:14:39.655: Error loading '/usr/lib/liferea/plugins/trayicon/trayicon.plugin'
> ```

I found an issue that was still causing this. Should be fixed with 93cf128a416f9aa7a3b31ce96e305e9f3862caf6

AndyM48:
I thought the trayicon.plugin file was redundant, a left over from the past versions?
I deleted it before reporting that the trayicon was working for me. (except for the icon itself which is a different issue)

lwindolf:
> I thought the trayicon.plugin file was redundant, a left over from the past versions? I deleted it before reporting that the trayicon was working for me. (except for the icon itself which is a different issue)

For the new C plugin a trayicon.plugin file is still needed. It is just compiled as a gresource and must not appear in the plugin directory anymore. In libpeas C plugins and Python plugins are totally different.

PR #1556, opened 2026-06-30
Node source refactoring

lwindolf:


lwindolf:
Current state: needs further testing, especially item state fetching and initial feed list updating.

lwindolf:
Arghh... recent changes drop all items on reload

lwindolf:
Refactoring is now stable and does not drop items anymore.

PR #1574, opened 2026-07-31
Feedlist to gtklistview

lwindolf:
Open problems

- [x] performance lag after marking item read ~300ms on my system
- [x] check for unnecessary feed list rebuilds 
- [ ] get DnD working without crashes

### DIFF
--- a/resources/ui/gtk-liferea.css
+++ b/resources/ui/gtk-liferea.css
@@ -7,3 +7,23 @@ frame.ruleEditor {
   border-radius: 0;
   padding: 12px;
 }
+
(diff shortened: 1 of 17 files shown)


lwindolf:
Currently excessively slow when skimming through unread items due to recursive unread count updates

lwindolf:
Unread count update cost has now been improved.

lwindolf:
Problem was with using GtkListBox instead of GtkListView

lwindolf:
Current status: DnD actions crash, because some row references or so are removed to early.

lwindolf:
Merging despite the DnD crashiness.

PR #1575, opened 2026-08-01
Serialize feed item document as HTML, not XML

psimonyi:
Liferea extracts a document from each feed item and then serializes it to give to WebKit for rendering. However, it was serializing the document as XML (XHTML) while delivering it to the WebKit web view as HTML. This means WebKit would not parse it as intended, since there are parsing differences between HTML and XML/XHTML.

In particular, the one I noticed is that libxml serializes empty tags as self-closing tags, but in HTML the XML self-closing tag syntax is ignored. For elements where a closing tag is optional or forbidden like `img`, this doesn't matter, but for tags like `iframe` or `script` where the element is usually empty but a closing tag is required, it means the tag is never closed in the HTML syntax and so the remainder of the document ends up being children of the `iframe`. Then DOMPurify deletes the `iframe` and its children, which means it's deleted everything after the first `iframe`.

If you look at a feed with `iframe`s such as embedded videos without this patch, the post body will be cut off at the first video.

Instead, Liferea should serialize the document in the language WebKit will use to parse it.

Although this kind of difference between parsers can often be a security issue, it happens before DOMPurify runs so I think it less likely in this case.

### DIFF
--- a/src/xml.c
+++ b/src/xml.c
@@ -30,6 +30,7 @@
 #include <libxml/parser.h>
 #include <libxml/entities.h>
 #include <libxml/HTMLparser.h>
+#include <libxml/HTMLtree.h>
 #include <libxml/xpath.h>
 
 #include "common.h"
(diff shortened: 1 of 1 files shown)


lwindolf:
@psimonyi First thanks for the patch and taking the time to debug this!

Do you have a test feed with e.g. iframes I could use to test the fix?

psimonyi:
https://www.jefftk.com/news.rss has a recent post "Organ Pedals for Drumming" with a couple of embedded YouTube videos. (Sorry, I should have thought to include a test case in the first place.)

lwindolf:
Thanks for the test feed. Could reproduce the bug and that the fix works!

Issue #1588, opened 2026-08-21
Feed updates stuck on 2.0, even after restart


z1atk0:
## Version
2.0

## Problematic Behavior
The automatic feed updates somehow got stuck after some time (a few hours, maybe?). Right-clicking the feed and selecting "Update" only resulted in a popup "Feed is already being updated!", for each and every feed I tried. Even restarting `lifearea` did _not_ fix the problem, feeds continued to not being updated anymore, and could also not be updated manually (same popup).

Going back to 1.16.14, all feeds were immediately updated right away on startup, just as expected.

## Steps to Reproduce

1. upgrade from 1.16 to 2.0
2. wait a few hours until feeds don't seem to get updates anymore
3. have a look at the feed URLs to verify that there _are_ actually updates available
4. try to update the feed manually, or restart `liferea`
5. still no updates! 🙁 

## Expected Behavior
Feeds should update automatically as usual. If there is really an update ongoing in the background, it would be nice to have a possibility to force-quit the running update process and force-restart it. Although I doubt that this is the problem here, as even a `liferea` restart didn't fix the problem for me.

lwindolf:
Could you provide an screenshot of the update monitor in this state?

z1atk0:
Okay, I'll give 2.0 another try and upload a screenshot if that happens again. 👍

z1atk0:
Well, that didn't take too long. 🙂 If I mouseover the violet boxes, all of them say "age=(between 9500 and 10500), interval=3600", and nothing gets updated. All entries in the "Fetching" table show "State: 0" and "Flags: 0".

<img width="1382" height="951" alt="Image" src="https://github.com/user-attachments/assets/fd932d57-144e-493c-a6f7-2c5201882c7d" />


For example, here's the "wunschliste.de" feed as currently seen in `liferea`:

<img width="1910" height="992" alt="Image" src="https://github.com/user-attachments/assets/d016f00d-eba7-4e1c-bb25-4191fcea922b" />


If you check the feed source URL at https://www.wunschliste.de/xml/atom.pl, you'll see that the latest 9 items are missing (I've collapsed them in the screenshot so it's easier to match with the most recent item in `liferea`):

<img width="1901" height="941" alt="Image" src="https://github.com/user-attachments/assets/71a44f2e-c800-4f24-bcaa-42498522fda9" />

[...]

PR #1591, opened 2026-08-24
Rather forcefully initialize networking before updating feeds

moench-tegeder:
Since version 2.0 (but not in 2.0-RC4), liferea may on startup enter a state where the update monitor shows all feeds in "updating" state and libsoup3 keeps complaining about

````
soup_session_send_and_read_async: assertion 'SOUP_IS_SESSION (session)' failed"
````

with the feeds actually never receiving an update. See https://github.com/lwindolf/liferea/issues/1588 for what I believe to be a manifestation of this.

Before commit f6d9ecb8980f630b30508b6ea094ff267d7fe57d, this line https://github.com/lwindolf/liferea/blob/v2.0-RC4/src/feedlist.c#L268 guaranteed that the NetworkMonitor - and with it, the networking subsystem containing the soup sessions - was initalized before the updater actually ran.
After that change, I don't see how the code guarantees the NetworkMonitor to be initialized in time for the updater (but that might be as well on me, as I'm not that familiar with gtk applications). The UpdateQueues seem to assume that the soup sessions are valid from the get-go, and fail miserably if that assumption does nbot hold.

As a quick fix I added a a call to network_init() right after conf_init() - that fixes the issue for me, but is perhaps not the way this code was designed.

### DIFF
--- a/src/liferea_application.c
+++ b/src/liferea_application.c
@@ -143,6 +143,7 @@ on_app_startup (GApplication *gapp, gpointer user_data)
 	/* Configuration necessary for network options, so it
 	   has to be initialized before network_init() */
 	conf_init ();
+	network_init();
 
 	/* Setup update queue handling */
 	app->updateQueue = update_job_queue_get_instance ();


lwindolf:
Thank you for the fix! This should fix #1588

PR #1595, opened 2026-09-08
Fix prefs dialog changing the update interval when opened

psimonyi:
(I started writing this as a bug report, then figured out how to fix it in a fairly straightforward way.  So it turned into a PR.)

Opening the Preferences dialog resets the default feed update interval to "1" of the selected unit (1 hour, 1 day, or 1 minute, whichever unit the previous preference value would have been displayed as).

The dialog does not *show* the reset value until you close it and open it again.

**Steps to reproduce:**

0. In a terminal, run `gsettings monitor net.sf.liferea` (not required, but useful for debugging)
1. Open Preferences, and change the Default Feed Refresh Interval to 8 hours (or any value that is not 1 unit).  Close the dialog.
2. Open Preferences again and change nothing.  Close the dialog.
3. Open Preferences again.

**Expected results:**

In steps 2 and 3, the dialog should show 8 hours (the same value that was set in step 1).  The value in gsettings should also stay the same as what you set in step 1.

**Actual results:**

In step 2, the dialog shows the value from step 1, but `gsettings monitor` shows that the stored value changed when the dialog opened.

In step 3, the dialog shows the value reset to 1 hour.

**Problem:**

Using `preferences_dialog_setup_drop_down` sets up a callback to save the interval preference before the dropdown widget's value has been set properly, and then the dropdown's value is set, triggering the callback, later during the setup in `preferences_dialog_init` while the dropdown and spinbutton widgets' state is inconsistent.

<details>
In preferences_dialog.c, `preferences_dialog_init` calls `preferences_dialog_setup_drop_down` to fill the unit options list, then it sets the selected dropdown item according to the value of the `DEFAULT_UPDATE_INTERVAL` preference from gsettings.  However, `preferences_dialog_setup_drop_down` connected the `selected` notify signal to `on_drop_down_changed`, which means that callback runs when the dropdown value is being set in `preferences_dialog_init`.  At that point, the value in the spin button is 1 because that's what's in the UI definition and the code to set its value from the preference hasn't been reached yet.  And `on_drop_down_changed` then calls `on_default_update_interval_value_changed` which saves the value from the current widgets' (inconsistent) state to gsettings, just before we finally set the value of the spin button.
</details>

This can be fixed by setting the dropdown's value after setting the spinbutton's value so that the callback sees a consistent widget state.  This is the solution I used in this patch.

An alternative is to make sure the dropdown's callback isn't connected until after its initial value is set.  That would require more extensive changes, but might end up being nicer code overall.  Using `preferences_dialog_setup_drop_down` for the units doesn't seem like it quite fits the function's purpose, so you might be able to improve this.


lwindolf:
Reproduced

lwindolf:
Thank you for contributing this fix!

PR #1596, opened 2026-09-11
Refactor update interval handling + test cases

lwindolf:
- Isolate update interval logic into subscription_get_effective_interval()
- Add test cases for different scenarios
- New run.sh wrapper for test cases to allow using conf_init() in an isolated DBUS session


lwindolf:
Solution so far is buggy, never stops updating at all. Auto update logic is not yet correct.

lwindolf:
Problem is fixed now. Solution should be complete. Will test for 2-3 days

PR #1598, opened 2026-09-14
Don't delete elements that are potential footnotes

psimonyi:
When an article's HTML has an element containing only a single element which in turn contains a single character, Liferea will delete the nested element.  However, `<sup><a href='#fn-1'>1</a></sup>` is a common pattern for footnotes.  As a result, Liferea strips the footnote markers from many feeds.

[Molly White's *Citation Needed*](https://newsletter.mollywhite.net/feed) is an example of a feed demonstrating this issue frequently.

While debugging, I changed this code to log all elements it was removing and browsed a bunch of articles in feeds I follow, and I came across quite a few cases where the removed element was a footnote or similar link, and one case where the removed element was a space between words (with extra markup probably caused by a WYSIWYG editor).  I did not come across any cases where it would have been okay to delete the element or where the element was actually empty or insignificant.

I'm not sure why "empty elements" need to be deleted; the comment says "to get rid of empty picture/video/iframe divs" but I'm not sure what problem they cause.  An empty div shouldn't take up any space, and even if they do, some marker showing where content was removed is probably a *good* thing.  The commit in which it was introduced (1c9464d) doesn't mention any more details.

I propose removing the code to delete these not-quite-empty elements.


lwindolf:
Thank you for your PR! I solved this a bit differently by limiting the empty tag deletion to media tags for which the removal was intended. Additionally a bit more details were added on the rationale. I believe this fixes your use case.

